PR #19, opened 2021-11-10
Created UART input logic and placeholder movement function declarations

chasesza:


firmware/arduino/chessboard/uart.ino
```diff
@@ -0,0 +1,62 @@
...
+                    move_straight(buffer[1] - 'a', buffer[2] - '0', buffer[3] - 'a', buffer[4] - '0');
+                    break;
+                case 1:
+                    //move_to_graveyard(int startRow, int startCol, char color, char type)
+                    move_to_graveyard(buffer[1] - 'a', buffer[2] - '0', buffer[3], buffer[4]);
+                    break;
+                case 2:
+                    //Args: piece color (b/w), kingside or queenside (-/0) 
+                    //castle(char color, char side)
+                    castle(buffer[1], buffer[3]);
```
nashirj, line 45: You should also have something here about queenside versus kingside castling

chasesza: Right now I'm just passing the - or 0 to the castle function that we're going to create in the future. 

I was leaning towards keeping the communications part of the code as simple as possible and handling queenside/kingside and white/black in the movement function.

Do you or @LSaber36 have a preference as to where it's handled?

LSaber36: I agree, we want to separate the logic from the moves if possible. That's my opinion at least.

LSaber36: Also, all of the ESP32 code to be on one file. It's okay to make changes to the main code as long as it's in your branch, where it's safe.

LSaber36: This code should focus on receiving the UART bytes first and foremost, with interpreting them after. The first step is storing the bytes in a byte array of a set size, which I believe we said was 6.

nashirj: I think putting the message parsing code in a separate place is a good idea. Also, I understand what you are doing now, thanks for clarifying.

firmware/arduino/chessboard/uart.ino
```diff
@@ -0,0 +1,62 @@
...
+            buffer[byteNum] = incomingByte;
+        }
+
+        //Check if the buffer is full, process input
+        if (byteNum == 4)
+        {
+            //Reset buffer position
+            byteNum = -1;
+
+            //Process input
```
nashirj, line 31: Per our other thread, here would be a good spot to call a helper function to parse the message. Something like `parseMessageFromPi(&buffer)` and maybe that can return a bool which you can use to tell whether the message was valid or not (maybe bytes got lost etc).

firmware/arduino/chessboard/uart.ino
```diff
@@ -0,0 +1,62 @@
...
+
+        //Reset buffer position
+        if (incomingByte == '~')
+        {
+            byteNum = 0;
+        }
+        //Add byte to buffer
+        else if (byteNum != -1)
+        {
+            buffer[byteNum] = incomingByte;
```
nashirj, line 22: This can be `buffer[byteNum++] = incomingByte` and then the last if statement can be removed

firmware/arduino/chessboard/uart.ino
nashirj: When you make the `castle` function, can you make it clear what you're doing? A simple comment saying exactly what you explained here would be perfect

nashirj (commented):
Looks really good thanks for getting this started

firmware/arduino/chessboard/uart.ino
chasesza: > Also, all of the ESP32 code to be on one file. It's okay to make changes to the main code as long as it's in your branch, where it's safe.

I think we're better off using multiple files to avoid merge conflicts. The Arduino IDE automatically combines them, and it won't cause any issues as long as we declare functions in the primary chessboard.ino file.

firmware/arduino/chessboard/uart.ino
chasesza: I just did this, but I don't want to mark the conversation resolved until you've had a chance to look at it.

nashirj: Looks great

firmware/arduino/chessboard/uart.ino
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+bool parse_message_from_pi(char * message)
+{
+    //Move types 0 and 3 are valid for the same range of values
+    if (message[0] == 0 || message[0] == 3)
+    {
+        //Check that the inputs are in a valid range
+        if (message[1] < 'a' || message[1] > 'h' || message[2] < '0' ||  message[2] > '8' || message[3] < 'a' || message[3] > 'h' || message[4] < '0' ||  message[4] > '8')
```
nashirj, line 44: I think here you want to check if the values at 2 and 4 are in range [1,8] inclusive rather than [0, 8]. Same comment below

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -59,4 +59,16 @@ void loop()
...
+// Function Declarations
+
+
+// Message Parsing
+bool parse_message_from_pi(char * buffer);
+
+// Piece Movement
+void move_straight(int startRow, int startCol, int endRow, int endCol);
+void move_to_graveyard(int startRow, int startCol, char color, char type);
+void castle(char color, char queenside_or_kingside);
```
nashirj, line 73: I like the parameter name, makes it more clear, thanks again

firmware/arduino/chessboard/uart.ino
nashirj: I didn't know you can do that! That's pretty cool. I think it will make things cleaner if we can have things in separate files; will lead to 1) fewer merge conflicts and 2) easier to understand code IMO

firmware/arduino/chessboard/uart.ino
chasesza: Good catch. Thanks!

nashirj:
LGTM, I am OK with you merging this in if @LSaber36 is ok with it

PR #21, opened 2021-11-12
Adds simple CNN example for MNIST dataset

JuddBE:
Basic CNN example with 1 input channel and some comments to help explain important steps.


nashirj:
Looks great, I'm fine with you merging it in

PR #26, opened 2021-11-14
added a few stepper setup functions

adenm-10:


LSaber36:
Aden and I updated the code to use motor arrays to group pins, use preprocessors, and make the code more efficient by reducing  redundancy and increasing code reusability.

PR #35, opened 2021-11-17
Firmware function updates and organization

adenm-10:


firmware/arduino/chessboard/uart.ino
```diff
@@ -1,84 +1,92 @@
 char incomingByte;
 char  buffer[5];
```
LSaber36, line 2: there's an extra space between char and buffer

firmware/arduino/chessboard/uart.ino
```diff
@@ -1,84 +1,92 @@
...
             parseMessageFromPi(buffer);
         }
     }
 }
 
 bool parseMessageFromPi(char * message)
 {
-    //Move types 0 and 3 are valid for the same range of values
+    // Move types 0 and 3 are valid for the same range of values
     if (message[0] == 0 || message[0] == 3)
```
LSaber36, line 41: So this is a change I want you to make for all of these, but when you see any comparator operators, like ==, && and ||, use two spaces between the operator and the arguments. For example...
Good  if (boolean1  &&  boolean2)
Bad     if (boolean1 && boolean2)

LSaber36: It just makes it slightly easier to read

firmware/arduino/chessboard/uart.ino
```diff
@@ -1,84 +1,92 @@
...
 
     }
     else if (message[0] == 1)
     {
-        //Check validity for the first 3 bytes
-        if (message[1] < 'a' || message[1] > 'h' || message[2] < '1' ||  message[2] > '8' || ((message[3] != 'b') && message[3] != 'w'))
+        // Check validity for the first 3 bytes
+        if (message[1] < 'a' || message[1] > 'h' 
+           || message[2] < '1' ||  message[2] > '8' 
+           || ((message[3] != 'b') && message[3] != 'w'))
```
LSaber36, line 65: There are two extra sets of parenthesis here. They should be removed since the arguments are the same format as the rest of the comparisons and none of the other ones have these wack parenthesis.

firmware/arduino/chessboard/uart.ino
```diff
@@ -1,84 +1,92 @@
...
         moveToGraveyard(message[1] - 'a', message[2] - '1', message[3], message[4]);        
     }
     else if (message[0] == 2)
     {
-        //Check input validity
-        if (!(message[1] == 'b' || message[1] == 'w') || message[2] != '0' || message[4] != '0' || !(message[3] == '0' || message[3] == '-'))
+        // Check input validity
+        if (!(message[1] == 'b' || message[1] == 'w') 
+           || message[2] != '0' || message[4] != '0' 
+           || !(message[3] == '0' || message[3] == '-'))
```
LSaber36, line 81: @chasesza why did you do another negation here? Would it make less sense to move that out of the initial negation? It's up to you but I'm suggesting turing !(a  ||  !(b  ||  c)) into !a  &&  (b  ||  c)
Where a is that first set of stuff.

chasesza: I think it's more readable if we check that it's !( the valid options) for each character.

Also, there isn't a double not the way you listed. It's !(a || b) || c || d || !(e || f)

LSaber36: Ohhhhh, my bad, I misread that. Then this is all good.

nashirj: It might be easier to read if you do `(!a && !b)` for the two cases where you currently have `!(a || b)` to distinguish it from the other conditions you're checking. That way each condition is separated by an `||`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -25,6 +25,17 @@ void moveMotor(int motor[], int steps, int dir)
...
 
+void moveSpaces(int motor[], int spaces, int dir)
+{
+  int i;
+
+  digitalWrite(MOTOR_SLEEP, HIGH)
+  digitalWrite(MOTOR_RESET, HIGH)
+  digitalWrite(MOTOR_ENABLE, LOW)
+
+  for(i=0; i < )
```
LSaber36: So this isn't bad cuz I won't merge the PR until the code compiles and everything, but try to avoid committing code that will break compilation or doesn't make sense.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -29,13 +29,49 @@ void moveSpaces(int motor[], int spaces, int dir)
 {
   int i;
 
-  digitalWrite(MOTOR_SLEEP, HIGH)
-  digitalWrite(MOTOR_RESET, HIGH)
-  digitalWrite(MOTOR_ENABLE, LOW)
+// How many steps per space
+  float stepsPerSquare = 200; //subject to change, motor turns 1.8degrees/step and 200 steps/full revolution
```
LSaber36, line 33: For modularity, I would suggest making this a variable that's calculated once in setup. It would use #defines to do so. That way, we can change it easily, since it's a magic number rn. 
I imagine the calculation would be something like:
stepsPerSquare = MILLIMETERS_PER_SQUARE * STEPS_PER_MILIMETER;

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -29,13 +29,49 @@ void moveSpaces(int motor[], int spaces, int dir)
...
+  float numSteps = spaces * stepsPerSquare;
```
LSaber36, line 34: This should stay since it's unique to the function call.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -29,13 +29,49 @@ void moveSpaces(int motor[], int spaces, int dir)
 {
   int i;
 
-  digitalWrite(MOTOR_SLEEP, HIGH)
-  digitalWrite(MOTOR_RESET, HIGH)
-  digitalWrite(MOTOR_ENABLE, LOW)
+// How many steps per space
```
LSaber36, line 32: This comment should be on the same indent level as the code around it

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -29,13 +29,49 @@ void moveSpaces(int motor[], int spaces, int dir)
...
+  float numSteps = spaces * stepsPerSquare;
 
-  for(i=0; i < )
+  digitalWrite(MOTOR_SLEEP, HIGH);
+  digitalWrite(MOTOR_RESET, HIGH);
+  digitalWrite(MOTOR_ENABLE, LOW);
+
+  digitalWrite(motor[1], dir);
+
+  for(i=0; i < numSteps; i++);
```
LSaber36, line 42: This is nitpicky, but use i = 0 here

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -29,13 +29,49 @@ void moveSpaces(int motor[], int spaces, int dir)
...
+    digitalWrite(motor[0], HIGH);
+  }
 
 }
+
+void moveDiaganol(int motor[], int dir, int spaces)
+{
+  int i;
+
+  float stepsPerSquare = 200; //similary subject to change as above function dictation, same number as above though
```
LSaber36, line 55: Same as above

LSaber36, line 55: I would put this comment above this line since it's long, and it needs a space after the //

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -29,13 +29,49 @@ void moveSpaces(int motor[], int spaces, int dir)
...
+  float numSteps = spaces * stepsPerSquare;
+
+  digitalWrite(MOTOR_SLEEP, HIGH);
+  digitalWrite(MOTOR_RESET, HIGH);
+  digitalWrite(MOTOR_ENABLE, LOW);
+
+  digitalWrite(xMotor[1], dir);
+  digitalWrite(yMotor[1], dir);
+
+  for(i=0; i < numSteps; i++)
```
LSaber36, line 65: Same as above

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -36,6 +36,11 @@ int yMotor[4] = {MOTOR2_STEP_PIN, MOTOR2_DIR, MOTOR2_MS1, MOTOR2_MS2};
 #define Y_AXIS_ENDSTOP_SWITCH 32
 #define CHESS_TIMER_BUTTON 4
 
+// Distance Definitions
+#define MILLIMETERS_PER_SQUARE 200
+#define STEPS_PER_MILLIMETER 200
+float stepsPerSpace = MILLIMETERS_PER_SQUARE * STEPS_PER_MILLIMETER;
```
nashirj, line 42: I'll defer to @LSaber36 's opinion on this, but 1) stepsPerSpace should probably be constant since it doesn't change during the program, and 2) since it's a global constant maybe it should be named STEPS_PER_SPACE

LSaber36: It will be a constant, but I suggested this so we would have less calculations to do outside of the program. If we need to fine tune that, which I'm pretty sure we will, we'd need to calculate that outside of the program. That's why I suggested doing it this way. The cost is negligible since it's only one float and calculations only show up during setup()

nashirj: I just meant making it `const float STEPS_PER_SPACE` instead of what it is now! So it's clear in the program that it is not a variable

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -24,7 +25,54 @@ void moveMotor(int motor[], int steps, int dir) {
...
+  for(i = 0; i < numSteps; i++);
+  {
+    digitalWrite(motor[0], LOW);
+    delay(1);  // 1 milliSecond
+    digitalWrite(motor[0], HIGH);
+  }
+
+}
+
+void moveDiaganol(int motor[], int dir, int spaces)
```
LSaber36, line 50: should be moveDiagonal

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -36,6 +36,10 @@ int yMotor[4] = {MOTOR2_STEP_PIN, MOTOR2_DIR, MOTOR2_MS1, MOTOR2_MS2};
...
```
LSaber36, line 42: if you compile with arduino, it'll probably give you an error about this. I believe you need to split up the variable declaration and assignment. The assignment should be in setup if it doesn't work here.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -24,7 +25,54 @@ void moveMotor(int motor[], int steps, int dir) {
...
+{
+  int i;
+
+  // similary subject to change as above function dictation, same number as above though
+  float numSteps = spaces * stepsPerSquare;
```
LSaber36, line 55: should this be stepsPerSpace?

LSaber36: I believe this will also raise an error with arduino, since stepsPerSquare no longer exists.

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -36,6 +36,10 @@ int yMotor[4] = {MOTOR2_STEP_PIN, MOTOR2_DIR, MOTOR2_MS1, MOTOR2_MS2};
 #define Y_AXIS_ENDSTOP_SWITCH 32
 #define CHESS_TIMER_BUTTON 4
 
+// Distance Definitions
+#define MILLIMETERS_PER_SQUARE 200
+#define STEPS_PER_MILLIMETER 200
```
LSaber36, line 41: I believe this should be 5

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -36,6 +36,10 @@ int yMotor[4] = {MOTOR2_STEP_PIN, MOTOR2_DIR, MOTOR2_MS1, MOTOR2_MS2};
 #define Y_AXIS_ENDSTOP_SWITCH 32
 #define CHESS_TIMER_BUTTON 4
 
+// Distance Definitions
+#define MILLIMETERS_PER_SQUARE 200
```
LSaber36, line 40: this should be 63, since that's around how many steps are in 2.5 inches. We'll end up making the board squares based on millimeters anyway so it'll line up nicely.

PR #38, opened 2021-11-19
Restructured computer vision directory

nashirj:
This PR puts all jupyter notebooks in a centralized place and updated name of notebooks to better reflect what each notebook does. I didn't change any of the code.


nashirj:
little does he know I actually deleted the whole repository

ryan-msg

PR #43, opened 2021-11-20
Game loop v1

nashirj:


nashirj:
After talking with Lucas and Chase offline, we decided to update the OpCode to not have a move for castling, as this would require having the Arduino orchestrate two moves. Instead, we'll use two distinct move commands, one direct move without collision avoiding `SimpleMove` and one indirect move with collision avoiding `MovePieceAlongSquareEdges`

PR #48, opened 2021-11-23
updated index.html and script.js to provide a console output of the m…

RuthShryock:
…oves

### DIFF
--- a/web-app/web-frontend/script.js
+++ b/web-app/web-frontend/script.js
@@ -2,43 +2,19 @@
 // https://github.com/jhlywa/chess.js
 
 var board = null
-var $board = $('#myBoard')
 var game = new Chess()
-var squareToHighlight = null
-var squareClass = 'square-55d63'
-
-function removeHighlights (color) {
-  $board.find('.' + squareClass)
-    .removeClass('highlight-' + color)
-}
+var $status = $('#status')
+var $pgn = $('#pgn')
 
 function onDragStart (source, piece, position, orientation) {
   // do not pick up pieces if the game is over
   if (game.game_over()) return false
 
-  // only pick up pieces for White
-  if (piece.search(/^b/) !== -1) return false
-}
-
-function makeRandomMove () {
-  var possibleMoves = game.moves({
-    verbose: true
-  })
-
-  // game over
-  if (possibleMoves.length === 0) return
-
-  var randomIdx = Math.floor(Math.random() * possibleMoves.length)
-  var move = possibleMoves[randomIdx]
-  game.move(move.san)
-
-  // highlight black's move
-  removeHighlights('black')
-  $board.find('.square-' + move.from).addClass('highlight-black')
-  squareToHighlight = move.to
-
-  // update the board to the new position
-  board.position(game.fen())
+  // only pick up pieces for the side to move
+  if ((game.turn() === 'w' && piece.search(/^b/) !== -1) ||
+      (game.turn() === 'b' && piece.search(/^w/) !== -1)) {
+    return false
+  }
 }
 
 function onDrop (source, target) {
@@ -52,18 +28,7 @@ function onDrop (source, target) {
   // illegal move
   if (move === null) return 'snapback'
 
-  // highlight white's move
-  removeHighlights('white')
-  $board.find('.square-' + source).addClass('highlight-white')
-  $board.find('.square-' + target).addClass('highlight-white')
-
-  // make random move for black
-  window.setTimeout(makeRandomMove, 250)
-}
-
-function onMoveEnd () {
-  $board.find('.square-' + squareToHighlight)
-    .addClass('highlight-black')
+  updateStatus()
 }
 
 // update the board position after the piece snap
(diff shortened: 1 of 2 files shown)


web-app/web-frontend/script.js
```diff
@@ -72,12 +37,46 @@ function onSnapEnd () {
   board.position(game.fen())
 }
 
+function updateStatus () {
+  var status = ''
+
+  var moveColor = 'White'
+  if (game.turn() === 'b') {
+    moveColor = 'Black'
+  }
```
nashirj, line 46: It might be nice to use a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) here so it's clear what's going on. `var moveColor = (game.turn() == 'b') ? "Black" : "White";`

nashirj (commented):
The pgn looks great and I like the status message you constructed!

Can you please also add something that outputs the previous move in the form `e2e4`, where the move is `p1p2`, p1 being the starting square of the piece and p2 being the destination square? This is what we're going to pass to/from the board.

web-app/web-frontend/script.js
RuthShryock: Thank you! Ok I will work on that!

nashirj (changes requested):
Thanks for incorporating the changes!

I've attached a photo that shows three things that should change:

image

1) Instead of modifying the existing `move_to_san()` function, can you please make a new function called `move_to_gan()` (with a comment explaining what this function outputs) and put your modified version of `move_to_san()` in there?
2) When a piece is captured, we don't want to have the <piecetype>x (for example, in move two: should be 2. e4d5 d8d5 instead).
3) Don't need the letter indicating piece type (e.g., move three should be 3. f1c4 c8e6 instead)
4) When promoting, want something of the form `p1p2<lowercase-piece-type>` (e.g., instead of a7a8=Q, it would be a7a8q).

Thanks again for working on this!

web-app/web-frontend/script.js
nashirj: Thanks looks great!

nashirj:
To clarify: `move_to_gan()` stands for "grid algebraic notation" as opposed to "standard algebraic notation"

web-app/web-frontend/js/chess.js
```diff
@@ -732,6 +730,49 @@ var Chess = function (fen) {
...
+      var output = ''
+      if (move.flags & BITS.KSIDE_CASTLE) {
+        output = 'O-O'
+      } else if (move.flags & BITS.QSIDE_CASTLE) {
+        output = 'O-O-O'
+      } else {
+        if (move.piece !== PAWN) {
+          var disambiguator = get_disambiguator(move, moves)
+          output += disambiguator
+        }
```
nashirj, line 753: What does this code do/is it necessary? I think you may be able to remove it for simplicity's sake.

web-app/web-frontend/js/chess.js
```diff
@@ -732,6 +730,49 @@ var Chess = function (fen) {
...
+  
+      make_move(move)
+      if (in_check()) {
+        if (in_checkmate()) {
+          output += '#'
+        } else {
+          output += '+'
+        }
+      }
+      undo_move()
```
nashirj, line 770: I checked stockfish, and it gets confused when it receives moves with a `+` or `#` appended to the end. Can you please remove this part of the function?

web-app/web-frontend/js/chess.js
```diff
@@ -1504,6 +1545,7 @@ var Chess = function (fen) {
         }
   
         var moves = []
+        var pos = []
```
nashirj, line 1548: Is this variable used? If not, you can remove it.

web-app/web-frontend/js/chess.js
```diff
@@ -1526,11 +1568,11 @@ var Chess = function (fen) {
...
           }
-  
+          
           move_string =
-            move_string +
-            ' ' +
-            move_to_san(move, generate_moves({ legal: true }))
+          move_string +
+          ' ' + 
+          move_to_gan(move, generate_moves({ legal: true }))
```
nashirj, line 1575: Instead of modifying how the pgn is displayed/saved, can you please display the last move in GAN format in addition to the pgn code that you previously had on the page? i.e. please change this line back to `move_to_san` and add something to script.js and index.html to display the last move in GAN format. Thank you

web-app/web-frontend/script.js
```diff
@@ -72,12 +38,43 @@ function onSnapEnd () {
...
+    status = moveColor + ' to move'
+
+    // check?
+    if (game.in_check()) {
+      status += ', ' + moveColor + ' is in check'
+    }
+  }
+
+  $status.html(status)
+  $pgn.html(game.pgn())
```
nashirj, line 67: I am not super familiar with JS but I think this is where you can display the last move in GAN format.

nashirj (commented):
Lmk if any of the comments are unclear! Thanks again for working on this.

nashirj (approved):
Looks great, nice job!

PR #52, opened 2021-11-30
Mcts testing and AI neural network

JuddBE:
Adds monte carlo tree search example and the chess AI CNN architecture.


chess-AI/experimental-work/NN Layout.py
```diff
@@ -0,0 +1,83 @@
...
+
+# this is the basic AI CNN layout
+# we will use this and improve it until it can be used for the final product.
+
+
+import torch
+import numpy as np
+from torch import nn
+import torch.nn.functional as F
+import torchvision
```
nashirj, line 11: Can clean up these imports; 1) torchvision and numpy don't seem to be used 2) if you `import torch`, you don't need `from torch import nn`, can just do `torch.nn` when you need to access, same thing with `torch.nn.functional`. instead of `import torch`, maybe you can do `from torch import nn, randn`

chess-AI/experimental-work/NN Layout.py
```diff
@@ -0,0 +1,83 @@
+
+
+# this is the basic AI CNN layout
+# we will use this and improve it until it can be used for the final product.
```
nashirj, line 4: This should be a [module docstring](https://google.github.io/styleguide/pyguide.html#382-modules), and the docstring should be the first line in the file (no newlines at the top)

chess-AI/experimental-work/NN Layout.py
```diff
@@ -0,0 +1,83 @@
...
+        self.convLayer = nn.Sequential(nn.Conv2d(in_channels=119, out_channels=256, kernel_size=3, padding=1),
+                                       nn.BatchNorm2d(256),
+                                       nn.ReLU(inplace=True))
+        
+        # Each conv layer in a residual block uses 256 3x3 filters, padding used to keep the channel dimensions constant.
+        self.resBlock = nn.Sequential(nn.Conv2d(in_channels=256, out_channels=256, kernel_size=3, padding=1), 
+                                    nn.BatchNorm2d(256), 
+                                    nn.ReLU(inplace=True), 
+                                    nn.Conv2d(in_channels=256, out_channels=256, kernel_size=3, padding=1), 
+                                    nn.BatchNorm2d(256))
```
nashirj, line 35: Small formatting nitpick, but can you align all the `nn.*`? Will make it look nicer/easier to read

nashirj: Same comment for below

chess-AI/experimental-work/NN Layout.py
```diff
@@ -0,0 +1,83 @@
...
+                                       nn.Linear(in_features=256, out_features=1),
+                                       nn.Tanh())
+
+        
+
+    def forward(self, x):
+        x = self.convLayer(x)
+        
+        # Go through all the residual blocks and add the input of the block to the output before regularization.
+        for numRes in range(3):
```
nashirj, line 63: I take it 3 is the number of residual blocks? Another way to do this so that you don't have a magic number here is to define a class attribute `self.num_residual_blocks = 3`. Then your for loop can be
```python
for _ in range(self.num_residual_blocks):
    do_stuff()
```
This makes it clear that you are doing the same thing in every loop, not really using the index.

chess-AI/experimental-work/NN Layout.py
```diff
@@ -0,0 +1,83 @@
...
+        return (policy_out, value_out)
+    
+    
+    
+model = PlayNetwork()
+policy, value = model(torch.randn(1, 119, 8, 8))
+policy = policy.reshape(73*8*8)
+value = value.item()
+print(policy)
+print(value)
```
nashirj, line 80: Can you please wrap this in a main function?

```python
def main():
    your_code()

if __name__ == '__main__':
    main()
```

nashirj (changes requested):
Hey @JuddBE I only reviewed your nn code and not the MTCS stuff, I'll look at that later. Some general feedback in addition to what I put in the file is to please use [snake case](https://en.wikipedia.org/wiki/Snake_case) instead of [camel case](https://en.wikipedia.org/wiki/Camel_case) for your variable names and to run [pylint](https://pylint.org/) on your code and try to address as much of the feedback it gives you as you can before uploading!

Otherwise, what you wrote looks great - thanks again for working on this

chess-AI/experimental-work/NN Layout.py
```diff
@@ -1,83 +1,77 @@
...
+                                        nn.BatchNorm2d(256),
+                                        nn.ReLU(inplace = True))
         
         # Each conv layer in a residual block uses 256 3x3 filters, padding used to keep the channel dimensions constant.
-        self.resBlock = nn.Sequential(nn.Conv2d(in_channels=256, out_channels=256, kernel_size=3, padding=1), 
-                                    nn.BatchNorm2d(256), 
-                                    nn.ReLU(inplace=True), 
-                                    nn.Conv2d(in_channels=256, out_channels=256, kernel_size=3, padding=1), 
-                                    nn.BatchNorm2d(256))
+        self.res_block = nn.Sequential(nn.Conv2d(in_channels = 256, out_channels = 256, kernel_size = 3, padding = 1, bias = False),
```
nashirj, line 25: From style guide: "Never use spaces around `=` when passing keyword arguments or defining a default parameter value, with one exception: when a type annotation is present, do use spaces around the `=` for the default parameter value." ([link to whitespace stuff](https://google.github.io/styleguide/pyguide.html#36-whitespace)). So basically you would instead have `in_channels=256` etc like you had before and similar everywhere else.

chess-AI/experimental-work/NN Layout.py
```diff
@@ -1,83 +1,77 @@
...
 
         
-
     def forward(self, x):
-        x = self.convLayer(x)
+        x = self.conv_layer(x)
         
         # Go through all the residual blocks and add the input of the block to the output before regularization.
-        for numRes in range(3):
+        for num_res in range(self.num_res_blocks):
```
nashirj, line 56: Since you don't use `num_res` in this loop, you can do `for _ in range(self.num_res_blocks)` to make it clear that the index variable is not used.

nashirj (commented):
Looks a lot better! Thanks for making those changes

nashirj:
One thing I just noticed is the name of the file. The [style guide talks about module naming](https://google.github.io/styleguide/pyguide.html#316-naming) - if you wanted to keep the filename basically the same following that convention you could change it to `nn_layout.py` or maybe `kg_nn.py`. Then we can do `from kg_nn import PlayNetwork` in other files once we get this integrated into the main game loop.

ryPattillo:
Merge MCTS into Master

PR #58, opened 2021-12-03
adding a popup with pawn promotion options

RuthShryock:
the code becomes stuck after I select a piece to promote the pawn to


nashirj:
I'm merging this since it's functional! You can keep working on the UI and make a new PR when you get the promotion popup thing looking nicer

PR #61, opened 2021-12-05
Turnwise input rep

### LINKED ISSUE #56: Add directionality to chess input representation
Before creating input to network, we need to have the board flipped to the person that's playing.

fshcat:
Flips board to the perspective of the current player and orders current player's pieces and castling planes first (closes #56). Also refactored for brevity.

### DIFF
--- a/chess-AI/experimental-work/state_representation.py
+++ b/chess-AI/experimental-work/state_representation.py
@@ -2,76 +2,57 @@ import torch
 import chess
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


nashirj (approved):
This is really great! Nice job. I left a couple comments but after you address them I'm fine with you merging it in. I realize this PR also addresses something we overlooked where the values of the planes were not being reset after each input. Thanks again for working on this.

PR #68, opened 2021-12-15
move diagonal and wrapper function update

adenm-10:


firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -27,39 +27,28 @@ void moveMotor(int motor[], int steps, int dir)
 
 void moveSpaces(int motor[], int spaces, int dir)
 {
-  int i;
-
   // How many steps per space
   float numSteps = spaces * stepsPerSpace;
```
LSaber36, line 31: Now that I'm thinking about it, this should be an int. We're not doing half of a step or anything.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -27,39 +27,28 @@ void moveMotor(int motor[], int steps, int dir)
...
   digitalWrite(motor[1], dir);
 
-  for(i = 0; i < numSteps; i++);
-  {
-    digitalWrite(motor[0], LOW);
-    delay(1);  // 1 milliSecond
-    digitalWrite(motor[0], HIGH);
-  }
-
+  moveMotor(motor[], (int)numSteps, dir)
```
LSaber36, line 35: With this declared as an int, you don't need to cast it as one explicitly

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -41,22 +41,42 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
   digitalWrite(MOTOR_RESET, HIGH);
   digitalWrite(MOTOR_ENABLE, LOW);
 
   digitalWrite(xMotor[1], dirX);
-  digitalWrite(yMotor[1], dirX);
-
-  for(i = 0; i < numSteps; i++)
+  digitalWrite(yMotor[1], dirY);
+  
+  if (numStepsX = numStepsY)
```
LSaber36, line 53: This should be ==, right now I't doing an assignment and always evaluating true cuz, for arduino, >0 is considered true

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -41,22 +41,42 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
-  digitalWrite(yMotor[1], dirX);
-
-  for(i = 0; i < numSteps; i++)
+  digitalWrite(yMotor[1], dirY);
+  
+  if (numStepsX = numStepsY) 
   {
+    for(i = 0; i < numStepsX; i++)
+    {
     digitalWrite(xMotor[0], LOW);
```
LSaber36, line 57: The lines in the loop should be indented by one

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -41,22 +41,42 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
     digitalWrite(yMotor[0], LOW);
     delay(1);
     digitalWrite(xMotor[0], HIGH);
     digitalWrite(yMotor[0], HIGH);
+    }
+  } 
+
+  else 
+  {
+    float dist = sqrt(numStepsX * numStepsX + numStepsY * numStepsY);
```
LSaber36, line 67: I see what you're doing here, but I don't think we need to do this. We just need to make the motors move an x and y distance.

LSaber36: Additionally, I talked to Nashir and we agreed that we don't need to worry about all slopes, just 1, 1/2, and 2. That being said, you want to set the motor scale's and drive them the same number of steps. With the scale different, one will do full steps and the other 1/2 steps. So the slope will be maintained. Or the other way around and that's a slope of 2.

LSaber36: To do this, use if statements to check which is bigger and the slope between them, which is just y/x

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -41,22 +41,42 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
+    float delayX = dist / numStepsX;
+    float delayY = dist / numStepsY;
+
+    for(i=0; i < dist; i++)
+      digitalWrite(xMotor[0], LOW);
+      //do i have to convert this into an int, can i keep it as a variable?
```
LSaber36, line 73: The delay here is just the time between pulses, so you can make that a preprocessor and use that, like we've done before for magic numbers.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -41,22 +41,42 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
+  } 
+
+  else 
+  {
+    float dist = sqrt(numStepsX * numStepsX + numStepsY * numStepsY);
+    float delayX = dist / numStepsX;
+    float delayY = dist / numStepsY;
+
+    for(i=0; i < dist; i++)
+      digitalWrite(xMotor[0], LOW);
```
LSaber36, line 72: For writing these low and high, try to group them together. So you should write both low, delay, them both high. Otherwise, you add extra delay and the movement may not be as smooth.

firmware/arduino/chessboard/motorDriver.ino
nashirj: Sorry for the noob question, but do we need to explicitly consider negative slopes as well?

firmware/arduino/chessboard/motorDriver.ino
nashirj: Apparently VS code has [auto formatting](https://stackoverflow.com/questions/29973357/how-do-you-format-code-in-visual-studio-code-vscode), hopefully this is useful for your future PRs, you can just auto format before pushing the code!

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -50,33 +80,45 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
       digitalWrite(xMotor[0], LOW);
-      //do i have to convert this into an int, can i keep it as a variable?
-      delay(delayX);
+      digitalWrite(yMotor[0], LOW);
+      delay(1);
       digitalWrite(xMotor[0], HIGH);
+      digitalWrite(yMotor[0], HIGH);
+    }
+  }
+  else if (numStepsY < numStepsX)
```
LSaber36, line 110: Just to be safe, I would add the same check for half like you did earlier.
Keep in mind, you have integer math here, so we can account for that by reversing x and y in the calculation.

&& (numStepsX / numStepsY) == 2

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -50,33 +80,45 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
     }
   } 
 
-  else 
+  else if (numStepsY > numStepsX && (numStepsY / numStepsX) == 2)
   {
-    float dist = sqrt(numStepsX * numStepsX + numStepsY * numStepsY);
-    float delayX = dist / numStepsX;
-    float delayY = dist / numStepsY;
+    //can I call setScale here even thought it's declared after?
```
LSaber36, line 97: From now on, ask any questions in the commit the same way I'm giving you feedback. That keeps the code free of unrelated comments. Also, if you make a comment please have a space after the //

firmware/arduino/chessboard/motorDriver.ino
LSaber36: If we don't have this we might end up in a situation where we're moving at a different slope.

LSaber36: However, we may end up just using numStepsY > numStepsX and an else provided we only call this function with diagonal moves that have those slopes. We can adjust that later with Nashir.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -50,33 +80,45 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
-    delay(1);
-    digitalWrite(xMotor[0], HIGH);
-    digitalWrite(yMotor[0], HIGH);
+      digitalWrite(xMotor[0], LOW);
+      digitalWrite(yMotor[0], LOW);
+      delay(1);
+      digitalWrite(xMotor[0], HIGH);
+      digitalWrite(yMotor[0], HIGH);
     }
   }
```
LSaber36, line 94: There shouldn't be a newline here

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -50,33 +80,45 @@ void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
...
+    for(i = 0; i < numStepsX; i++)
+    {
+      digitalWrite(xMotor[0], LOW);
       digitalWrite(yMotor[0], LOW);
-      delay(delayY);
+      delay(1);
+      digitalWrite(xMotor[0], HIGH);
       digitalWrite(yMotor[0], HIGH);
+    }
   }
```
LSaber36, line 122: I have one more comment about this code, but I want to wait until you push the next set of changes. It's about the repeated for loop. See if you can figure it out :)

LSaber36 (approved):

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -27,55 +27,12 @@ void moveMotor(int motor[], int steps, int dir)
...
-  }
-
-}
-
-void moveDiagonal(int motor[], int dir, int spaces)
-{
-  int i;
-
-  // similary subject to change as above function dictation, same number as above though
-  float numSteps = spaces * stepsPerSpace;
```
LSaber36, line 55: This should be an int

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -106,4 +63,50 @@ void setScale(int motor[], int scale)
...
   }
-}
\ No newline at end of file
+}
+
+void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
+{
+  int i;
+
+  float numStepsX = spacesX * stepsPerSpace;
```
LSaber36, line 72: These should be ints

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -106,4 +63,50 @@ void setScale(int motor[], int scale)
...
+ else if (numStepsY > numStepsX && (numStepsY / numStepsX) == 2)
+  {
+    setScale(xMotor, 2);
+    setScale(yMotor, 1);
+  }
+  else if (numStepsY < numStepsX && (numStepsX / numStepsY) == 2)
+  {
+    setScale(xMotor, 1);
+    setScale(yMotor, 2);
+  }
```
LSaber36, line 96: Should be a newline here

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -106,4 +63,50 @@ void setScale(int motor[], int scale)
...
+
+  digitalWrite(xMotor[1], dirX);
+  digitalWrite(yMotor[1], dirY);
+  
+  if (numStepsX == numStepsY)
+  {
+    setScale(xMotor, 1);
+    setScale(yMotor, 1);
+  }
+ else if (numStepsY > numStepsX && (numStepsY / numStepsX) == 2)
```
LSaber36, line 87: This should be indented properly

PR #69, opened 2021-12-16
Game loop v3

nashirj:
Big PR, sorry :/

- Created first draft of a move queue; this allows breaking moves up into constituent moves (e.g. for captures or castling) and sending each of these moves to the board individually.
- Added feature to convert chess coordinates (e.g. a1) to board coordinates (e.g. (2, 2)) so that when sending moves to Arduino, Arduino can just receive an "absolute" start and end point on the board. This conversion works whether the white pieces are on the human side, or the black pieces are on the human side (see `get_board_coords_from_square`).
- Updated chessboard indexing so that (0, 0) corresponds to a1.
- Updated board indexing so that (0, 0) corresponds to the bottom left corner (when viewing the board top down with the human at the "bottom").
- Updated ArduinoStatus class to contain a status, move count, and an extra bit. Also created a helper method to parse Arduino messages into those three components
- Created function to check if knight move has neighbors along path; used to determine whether it is possible to move the knight diagonally or whether we need to move piece along square edges.
- Created first draft of message sent over serial to Arduino to specify move to make.
- Updated all graveyard filling methods.
- Created Move class to wrap the moves that are being sent to the Arduino.

### DIFF
--- a/firmware/raspi/boardinterface.py
+++ b/firmware/raspi/boardinterface.py
@@ -16,8 +16,10 @@ class Engine:
     Attributes:
         board: A python wrapper around a python-chess board.
     '''
-    def __init__(self):
+    def __init__(self, human_plays_white_pieces):
         self.chess_board = chess.Board()
+
+        self.human_plays_white_pieces = human_plays_white_pieces
 
         self.king_to_rook_moves = {}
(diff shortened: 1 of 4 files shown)


JuddBE:
Looks good to merge, great progress!

nashirj:
Thanks so much!

PR #70, opened 2021-12-22
Update existing driver functions and create new ones

LSaber36:


firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -42,40 +89,68 @@ void moveSpaces(int motor[], int spaces, int dir)
...
-void moveDiagonal(int dirX, int dirY, int spacesX, int spacesY)
+void moveDiagonal(int startCol, int startRow, int endCol, int endRow)
 {
+  int dirX, dirY, spacesX, spacesY;
+  int numStepsX, numStepsY;
   int i;
 
-  int numStepsX = spacesX * stepsPerSpace;
-  int numStepsY = spacesY * stepsPerSpace;
+  // Abs ensures that numStepsx and numStepsY will be positive
```
nashirj, line 143: Should be `numStepsX`. Also, instead of saying what happens, comment should explain why you need them to be positive.
```c
// Get magnitude of numStepsX and numStepsY; direction of
// movement is controlled by using dirX and dirY to set DIR_PIN.
```

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -42,40 +89,68 @@ void moveSpaces(int motor[], int spaces, int dir)
...
-  else if (scale == 4)  // 1/4 steps
+  else if (endCol == startCol)
   {
-    digitalWrite(motor[2], LOW);
-    digitalWrite(motor[3], HIGH);
+    // Y movement
+    spaces = abs(endRow - startRow);
+    dir = (endRow > startRow) ? UP : DOWN;
+    setScale(yMotor, 1);
   }
```
nashirj, line 115: There is an unhandled case where endRow != startRow and endCol != startCol. I know this is an assumption of the function (and I think we already talked about this in Aden's PR), but if this gets called without either of those cases addressed, the program will have a garbage value when you do `numSteps = spaces * stepsPerSpace;` since `spaces` is uninitialized, which will cause weird behavior. Can simply have an `else` condition that raises an error or returns or something.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -42,40 +89,68 @@ void moveSpaces(int motor[], int spaces, int dir)
...
+
+  // Set direction of motor
+  digitalWrite(motor[1], dir);
+
+  // Rotate motor some number of steps
+  for (i = 0; i < numSteps; i++) 
   {
-    digitalWrite(motor[2], HIGH);
-    digitalWrite(motor[3], HIGH);
+    if (digitalRead(X_AXIS_ENDSTOP_SWITCH) == HIGH  ||  digitalRead(Y_AXIS_ENDSTOP_SWITCH))
```
nashirj, line 128: I believe this should be
```
if (digitalRead(X_AXIS_ENDSTOP_SWITCH) == HIGH ||
    digitalRead(Y_AXIS_ENDSTOP_SWITCH) == HIGH)`
```

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -42,40 +89,68 @@ void moveSpaces(int motor[], int spaces, int dir)
...
+      break;
```
nashirj, line 129: When breaking because we hit the endstop, should we do some error handling? I imagine we shouldn't hit the end stops when calling a move from location A to location B. Even if you don't handle this now, maybe you can add a TODO?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -4,36 +4,83 @@
...
 
-  // Rotate motor some number of steps
-  for (i = 0; i < steps; i++) 
+void homeAxis(int motor[])
+{
+  int i;
+
+  // Loop until endstop collision, then fine tune it
+  // Since we're reusing this code for the x and y axis, 
+  // LEFT and RIGHT don't always apply to the motor passed in, but the values are the same
```
nashirj, line 51: Alternate phrasing that is possibly more clear:
```
// LEFT and DOWN have same value, as do RIGHT and UP. Use LEFT
// and RIGHT arbitrarily while tuning end stop for both x and y axes.
```

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -39,7 +26,22 @@ int yMotor[4] = {MOTOR2_STEP_PIN, MOTOR2_DIR, MOTOR2_MS1, MOTOR2_MS2};
 // Distance Definitions
 #define MILLIMETERS_PER_SQUARE 63
 #define STEPS_PER_MILLIMETER 5
+#define HOME_OFFSET 100
```
nashirj, line 29: Sorry for the noob question, but what does this variable represent? Offset from bottom left corner of board?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -4,36 +4,83 @@
   0         1        2        3
   Step pin, Dir pin, MS1 pin, MS2 pin
 */
```
nashirj, line 6: Please update these comments describing pins explaining what each of the five elements of the motor array correspond to now that you've added the limit switch to the motor array (or simply remove the comments).

nashirj (commented):
Here are a few more general comments sort of related to this PR but could be addressed in future PR:
- Consider making a `Point` struct to make parameter passing less verbose rather than having `int`s everywhere. This would allow `void moveAlongEdges(Point* start, Point* end)` and similar. Point could be defined like
```
struct Point {
    int col;
    int row;
}
```
- Can you please add function docstrings explaining what the purpose of each function is? Here is a description of what kind of thing I mean from a C++ style guide:
> Almost every function declaration should have comments immediately preceding it that describe what the function does and how to use it. These comments may be omitted only if the function is simple and obvious (e.g., simple accessors for obvious properties of the class). Private methods and functions declared in `.cc` files are not exempt. Function comments should be written with an implied subject of This function and should start with the verb phrase; for example, "Opens the file", rather than "Open the file". In general, these comments do not describe how the function performs its task. Instead, that should be left to comments in the function definition.
- Question about endstops: is there only one per axis, or is there one on each end of the axis for a total of four?
- This isn't a specific comment to this PR but to avoid using magic numbers for the array indexing, maybe you can `#define STEP_PIN 1` and same for the other pins so that you can have something like `digitalWrite(motor[STEP_PIN])` which is more readable than using an integer index.
- As above, but w.r.t. scale: can have `#define WHOLE_STEPS 1`, `#define WHOLE_STEPS 2`, etc to get rid of magic numbers.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -42,40 +89,68 @@ void moveSpaces(int motor[], int spaces, int dir)
...
-void setScale(int motor[], int scale) 
+
+void moveStraight(int motor[], int startCol, int startRow, int endCol, int endRow)
```
nashirj, line 93: Just want to clarify that the messages I'm sending from Pi are in the form xy where x is the row and y is the column. So this swap of parameter ordering may be confusing and require extra care when calling `moveStraight()` to make sure we pass parameters correctly. Might be better to leave in the order it was previously. This comment is irrelevant if you convert the int parameters to a Point before calling this function.

firmware/arduino/chessboard/motorDriver.ino
nashirj: This is being addressed in #73

nashirj:
Looks great

PR #75, opened 2021-12-28
Update README

nashirj:


LSaber36:
Not sure we should mention the voice recognition. That might be a goal for later on possibly.

nashirj:
Updated wording to mention it as future work.

PR #77, opened 2022-01-06
Updated pin definitons

LSaber36:


nashirj (commented):
I’m reviewing on my phone so just dumping all comments in here. Also, I can’t speak to the correctness of the updated pin definitions since I don’t have the spec sheet on hand, but maybe @chasesza or @adenm-10 can review those.

It might be clearer to put shared motor pins after defining x and y pins
Comment at 49 should say “Motor step size definitions” maybe?
Should have comment at line 55, “Motor array index definitions” or something
Could group pin definitions more clearly with
// START: Pin definitions
// END: Pin definitions
since labeling the other section “Preprocessor definitions” is slightly inaccurate because they are all preprocessor defs. If u do it that way then you can just remove the Preprocessor defs comment
Line 41 should prob have a comment preceding it saying “Global program state variables” or something and placed after all other definitions

PR #82, opened 2022-01-13
Board state detection work

nashirj:
Prototype code for computer vision. Need hardware solution for different lighting conditions before we can finish implementing computer vision.


nashirj:
This can be merged; I think the notebooks will be just for prototyping stuff, since the "production" code is in the game loop now

PR #85, opened 2022-01-14
Add chess timer button input interrupt

chasesza:


firmware/arduino/chessboard/chessboard.ino
```diff
@@ -89,6 +89,8 @@ void setup()
   stepsPerSpace = MILLIMETERS_PER_SQUARE * STEPS_PER_MILLIMETER;
 
   Serial2.begin(115200, SERIAL_8N1, RX2, TX2);
+  
+  attachInterrupt(digitalPinToInterrupt(CHESS_TIMER_BUTTON), chess_timer_ISR, RISING);
```
chasesza, line 93: @LSaber36  Are we using active high or active low for the buttons?

firmware/arduino/chessboard/uart.ino
```diff
@@ -28,6 +30,20 @@ enum ErrorCode
     MOVEMENT_ERROR = '4'
 };
 
+// Send message to Pi when the chess timer is pressed
+void chess_timer_ISR()
```
chasesza, line 34: The capital ISR doesn't follow the style guide, but chess_timer_isr() seems wrong to me for some reason. I can change it if you'd prefer.

firmware/arduino/chessboard/uart.ino
```diff
@@ -70,7 +95,14 @@ void serialEvent2()
                 // Sends acknowledgement
                 sendMessageToPi(currentState, buffer[5], errorCode);
 
+                movementFlag = true;
                 makeMove(buffer);
+
+                // Tell Pi that the chess timer button was pressed
+                if (transmitFlag)
+                    sendMessageToPi(END_TURN, 0, buffer[5]);
+                movementFlag = false;
```
nashirj, line 104: Can you please put this flag immediately after `makeMove` just for clarity's sake, ty

nashirj (approved):

firmware/arduino/chessboard/uart.ino
LSaber36: Please add a space between lines 103 and 104

firmware/arduino/chessboard/uart.ino
```diff
@@ -3,6 +3,10 @@ char buffer[6];
 int byteNum = -1; // -1 indicates that the start code hasn't been received
 char currentState = '0';
 char errorCode;
+volatile unsigned long previous_activation_time = 0;
+const unsigned long DEBOUNCE_TIME = 100; // Milliseconds
+bool movementFlag = false;
+volatile bool transmitFlag = false;
```
LSaber36, line 9: Why is one volatile and the other isn't?

firmware/arduino/chessboard/uart.ino
```diff
@@ -3,6 +3,10 @@ char buffer[6];
 int byteNum = -1; // -1 indicates that the start code hasn't been received
 char currentState = '0';
 char errorCode;
+volatile unsigned long previous_activation_time = 0;
+const unsigned long DEBOUNCE_TIME = 100; // Milliseconds
```
LSaber36, line 7: You could save memory by using a preprocessor in the main code

firmware/arduino/chessboard/uart.ino
```diff
@@ -3,6 +3,10 @@ char buffer[6];
...
+bool movementFlag = false;
+volatile bool transmitFlag = false;
 
 enum ArduinoState
```
LSaber36, line 11: @nashirj I'm not sure about this, but wouldn't it be better to store these in preprocessors since they wouldn't take up memory?

firmware/arduino/chessboard/uart.ino
nashirj: From my quick reading (and limited understanding) the volatile flag is only supposed to be used for variables that are modified in different threads. So only the two variables that are modified in the interrupt should have volatile flag, which Chase did here

firmware/arduino/chessboard/uart.ino
nashirj: Agree with making DEBOUNCE a define just for consistency with other constants

firmware/arduino/chessboard/uart.ino
nashirj: I think the variables are fine as is, but the const can be changed to a define

firmware/arduino/chessboard/uart.ino
chasesza: I did it wrong. I forgot that serialEvent2() is also a separate thread.

nashirj: Ah kk good catch Lucas

firmware/arduino/chessboard/chessboard.ino
LSaber36: Active high

firmware/arduino/chessboard/uart.ino
chasesza: I think it's grouped appropriately now that movementFlag is with the makeMove() function call.

chasesza:
```
movementFlag = true;
makeMove(buffer);
movementFlag = false;

// Tell Pi that the chess timer button was pressed
if (transmitFlag)
    sendMessageToPi(END_TURN, 0, buffer[5]);
transmitFlag = false;
```

Issue #102, opened 2022-01-19
Develop piece color classification capability


nashirj:
The options we have for developing this functionality are clearly laid out in these pages on the Raspberry Turk website:
1) http://www.raspberryturk.com/notebooks/chess_piece_presence.html
2) http://www.raspberryturk.com/details/vision/data_collection.html

nashirj:
Make a Jupyter notebook recreating results for raspberry turk dataset, then once we have a dataset with our board/camera config, just plug and play with the new images to develop the classifier.

Issue #101, opened 2022-01-19
Develop piece type classifier


nashirj:
Need a piece type classifier so that we can identify which piece type a piece is promoted to.

nashirj:
http://www.raspberryturk.com/notebooks/chess_piece_classification.html

nashirj:
While we are waiting for this functionality, can have user input to specify piece being promoted to.

Issue #100, opened 2022-01-19
Figure out how to detect corners of board


nashirj:
We have three main options:
1) Hard code the image corners into our game loop code based on inspection
2) Make the corners distinct colors and identify them with CV
3) Use some 3rd approach to detect corners (e.g. Harris Corner detector)

My preference is option 2, but we need to discuss with hardware team if we're doing that approach.

ColonelToad:
Would using a Harris Corner not be the most accurate way to detect corners? Or is there another reason it's not necessarily the preferred option?

nashirj:
![image](https://user-images.githubusercontent.com/20624069/152254924-2e90d117-1c8a-4e29-a37e-3ddf9a71e264.png)
![image](https://user-images.githubusercontent.com/20624069/152254956-59a2cd55-ad57-410f-93d2-b85545745507.png)

ColonelToad:
So you can see the results of running all the tests, overall Canny has the best accuracy however I was surprised to see Shi-Tomasi spotted pieces off the board as well. 
![Canny](https://user-images.githubusercontent.com/74569374/153775311-9e8a36c9-a523-40bb-811a-6df5400a22e5.png)
![Shi-Tomasi](https://user-images.githubusercontent.com/74569374/153775312-e43ec96c-2ac4-48c4-9479-1dc0de1f2524.png)
![Sobel](https://user-images.githubusercontent.com/74569374/153775314-c53955b7-cedd-4bb5-8961-4a34a1ac2ff9.png)

nashirj:
This looks great! Nice progress

ColonelToad:
So when using the new board image I'm getting much different results. I'll play around with the values more but Shi-Tomasi is not giving me anything at all. Canny is showing some of the edges on the area of the board where the board isn't under the light, and the gradients from the sobel so the same thing. I actually tried Harris and got some results with it though.
![canny](https://user-images.githubusercontent.com/74569374/153794635-c9c530b5-8fb4-4dd5-95dc-cd649dde355c.png)
![sobel](https://user-images.githubusercontent.com/74569374/153794637-39b5d3bb-a477-450d-adfb-d1a3d15811fb.png)
![harris](https://user-images.githubusercontent.com/74569374/153794638-1f304e4b-31d4-410c-97fb-b5cbb73794bf.png)

ColonelToad:
[...]

PR #103, opened 2022-01-19
Game loop cv

nashirj:
This PR sets up the basic code infrastructure for computer vision. We still need to be able to detect board corners (#100), classify piece type for promotion (#101), and piece occupancy for general board state detection (#102)

### DIFF
--- a/firmware/raspi/boardinterface.py
+++ b/firmware/raspi/boardinterface.py
@@ -56,7 +56,6 @@ class Engine:
         return self.chess_board.fen()
 
     @staticmethod
-    # TODO: replace this with a call to util.is_promotion(). Need previous_board_fen available.
     def is_promotion(uci_move):
         '''Returns boolean indicating if uci_move is a promotion.
         '''
--- a/firmware/raspi/boardinterface.py
+++ b/firmware/raspi/boardinterface.py
@@ -93,20 +93,20 @@ class Engine:
         '''
         return self.chess_board.is_capture(self.chess_board.parse_uci(uci_move))
 
-    @staticmethod
-    def get_chess_coords_from_square(square):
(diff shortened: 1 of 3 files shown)


nashirj:
Just realized there are a few things in boardinterface that need to be cleaned up still. I'll message when I'm ready for you to take a look

firmware/raspi/boardinterface.py
```diff
@@ -73,16 +73,6 @@ def is_capture(self, uci_move):
...
-    # def get_chess_coords_from_square(square):
-    #     '''Converts chess square to a BoardCell.
-
-    #     Example: a1 <=> [0, 0], h8 <=> [7, 7], regardless of whether human plays white or black
-    #     pieces.
-    #     '''
-    #     # Nums correspond to row (rank), letters correspond to col (files)
-    #     return util.BoardCell(ord(square[1]) - ord('1'), ord(square[0]) - ord('a'))
-
     def get_board_coords_from_square(self, square):
```
ColonelToad, line 76: Why did you decide get_chess_coords_from_square should be moved at the end of the uti.py file?

nashirj: I see now that I didn't need to. My thinking was that I didn't want to do `from boardinterface import Engine` in the `vision` file. Would you prefer I do that instead? I can change it back.

ColonelToad: No, I think that was a good change and was just wondering why.

nashirj: Sounds good!

JuddBE (approved):
Looks good to me!

nashirj:
Thanks for the review!

PR #109, opened 2022-01-20
doubled resolution of board by dividing MILLIMETERS_PER_SQUARE by 2

adenm-10:
I feel like I'm missing something but wouldn't just doing this double the amount of spaces in the board by halving the amount of steps needed per space? Since everythings magic numbers wouldn't that update everything?

### DIFF
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -38,7 +38,7 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 // ================================
 
 // Distance definitions
-#define MILLIMETERS_PER_SQUARE 63
+#define MILLIMETERS_PER_SQUARE 32
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
 float stepsPerSpace;
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -38,8 +38,7 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 // ================================
 
 // Distance definitions
-#define MILLIMETERS_PER_SQUARE 63
-#define MILLIMETERS_PER_SPACE 32
+#define MILLIMETERS_PER_UNITSPACE 32
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
 float stepsPerSpace;
@@ -91,8 +90,7 @@ void setup()
   pinMode(Y_AXIS_ENDSTOP_SWITCH, INPUT);
   pinMode(CHESS_TIMER_BUTTON, INPUT);
 
-  stepsPerSpace = MILLIMETERS_PER_SPACE * STEPS_PER_MILLIMETER;
-  stepsPerSquare = MILLIMETERS_PER_SQUARE * STEPS_PER_MILLIMETER
+  stepsPerSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
 
   Serial2.begin(115200, SERIAL_8N1, RX2, TX2);
   
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -41,8 +41,7 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 #define MILLIMETERS_PER_UNITSPACE 32
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
-float stepsPerSpace;
-float stepsPerSquare;
+float stepsPerUnitSpace;
 int currentX, currentY;
 
 // Motor directions
@@ -90,7 +89,7 @@ void setup()
   pinMode(Y_AXIS_ENDSTOP_SWITCH, INPUT);
   pinMode(CHESS_TIMER_BUTTON, INPUT);
 
-  stepsPerSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
+  stepsPerUnitSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
 
   Serial2.begin(115200, SERIAL_8N1, RX2, TX2);
   
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -38,10 +38,12 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 // ================================
 
 // Distance definitions
-#define MILLIMETERS_PER_SQUARE 32
+#define MILLIMETERS_PER_SQUARE 63
+#define MILLIMETERS_PER_SPACE 32
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
 float stepsPerSpace;
+float stepsPerSquare;
 int currentX, currentY;
 
 // Motor directions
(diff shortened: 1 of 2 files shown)


firmware/arduino/chessboard/chessboard.ino
```diff
@@ -38,7 +38,7 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 // ================================
 
 // Distance definitions
-#define MILLIMETERS_PER_SQUARE 63
+#define MILLIMETERS_PER_SQUARE 32
```
LSaber36, line 41: So, this is definitely the first step in the process of implementing this. But ask yourself, what's the effect of doing this with the current code? The answer is that is just reduces the size of each space. However, in this new scheme, one whole piece space will be two unit spaces wide and tall. So, with that in mind, how can we change the code to make one row or col equal two unit spaces?

adenm-10:
So here I made a difference between a space and a square and just multiplied all instances of spaces being calculated in functions by 2, with the sizes of spaces being halved shouldn't this adjust their rows and columns sizes appropriately?

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -39,9 +39,11 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 
 // Distance definitions
 #define MILLIMETERS_PER_SQUARE 63
+#define MILLIMETERS_PER_SPACE 32
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
 float stepsPerSpace;
+float stepsPerSquare;
```
LSaber36, line 46: Remove pls :)

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -89,7 +91,8 @@ void setup()
   pinMode(Y_AXIS_ENDSTOP_SWITCH, INPUT);
   pinMode(CHESS_TIMER_BUTTON, INPUT);
 
-  stepsPerSpace = MILLIMETERS_PER_SQUARE * STEPS_PER_MILLIMETER;
+  stepsPerSpace = MILLIMETERS_PER_SPACE * STEPS_PER_MILLIMETER;
+  stepsPerSquare = MILLIMETERS_PER_SQUARE * STEPS_PER_MILLIMETER
```
LSaber36, line 95: Remove pls :)

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -39,9 +39,11 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 
 // Distance definitions
 #define MILLIMETERS_PER_SQUARE 63
+#define MILLIMETERS_PER_SPACE 32
```
LSaber36, line 42: MILLIMETERS_PER_UNIT

adenm-10:
Was a little unsure what the goal of separating the point and integer but I applied it to numSteps, if that wasn't the intention just let me know and I'll adjust.

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -38,8 +38,7 @@ int yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS_END
 // ================================
 
 // Distance definitions
-#define MILLIMETERS_PER_SQUARE 63
-#define MILLIMETERS_PER_SPACE 32
+#define MILLIMETERS_PER_UNITSPACE 32
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
 float stepsPerSpace;
```
LSaber36, line 44: Only one of these is used and the name should have unit in it, like the preprocessor.

LSaber36: You can use ctrl+f to figure out which name only shows up once (which is during initialization.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -95,14 +95,14 @@ bool moveStraight(int motor[], int startCol, int startRow, int endCol, int endRo
   if (endRow == startRow)
   {
     // X movement
-    spaces = abs(endCol - startCol);
+    spaces = 2 * abs(endCol - startCol);
```
LSaber36: Remove these 2 *, there should only be a 2 * at the steps calculation.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -111,7 +112,7 @@ bool moveStraight(int motor[], int startCol, int startRow, int endCol, int endRo
     return false;
   }
 
-  numSteps = spaces * stepsPerSpace;
+  numSteps = 2 * pieceSpaces * stepsPerSpace;
```
LSaber36: This will give you an error if you compile the code. Since pieceSpaces is a float.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -137,22 +138,23 @@ bool moveStraight(int motor[], int startCol, int startRow, int endCol, int endRo
...
+  pieceSpacesX = abs(endCol - startCol);
   dirX = (endCol > startCol) ? RIGHT : LEFT;
 
-  spacesY = 2 * abs(endRow - startRow);
+  pieceSpacesY = abs(endRow - startRow);
   dirY = (endRow > startRow) ? UP : DOWN;
 
-  numStepsX = spacesX * stepsPerSpace;
-  numStepsY = spacesY * stepsPerSpace;
+  numStepsX = 2 * pieceSpacesX * stepsPerSpace;
```
LSaber36: Same float error as above.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -92,6 +92,16 @@ bool moveStraight(int motor[], float startCol, float startRow, float endCol, flo
   int dir, numSteps;
   int i;
 
+  //makes sure all arguments are in whole unitSpaces
```
LSaber36, line 95: Need a space after the // and capitalize the 'm'

LSaber36: Might want to reword this slightly, since whole and .5 may be confusing to someone reading the code. Maybe mention that  it makes sure all args can produce a multiple of unitspaces?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -112,7 +122,10 @@ bool moveStraight(int motor[], float startCol, float startRow, float endCol, flo
     return false;
   }
 
-  numSteps = 2 * pieceSpaces * stepsPerSpace;
+  if (pieceSpaces > floor(pieceSpaces))
+    numSteps = 2 * floor(pieceSpaces) * stepsPerSpace + stepsPerSpace
```
LSaber36, line 126: floor() returns a float, so you'll need to explicitly cast as an (int)

LSaber36: Also, this gets executed regardless of the if statement, so it should be above and outside of this if statement. The idea here is that we're calculating the total spaces, then adding a half space if necessary. That being said, you only do line 128 if there is a .5 present. That's why we separate the float into whole and decimal components.

Also, line 128 only adds half a unit square's number of steps, so it would be a plus-equal operation.

LSaber36: Make sure you compile your code before committing, it'll help show some of this stuff.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -83,26 +83,37 @@ void home()
...
+  // Makes sure all arugments produce a multiple of unitspaces
+  if (fmodf(startCol, 0.5) != 0)
+    return false;
+  else if (fmodf(startRow, 0.5) != 0)
+    return false;
+  else if (fmodf(endCol, 0.5) != 0)
+    return false;
+  else if (fmodf(endRow, 0.5) != 0)
+    return false;
+
```
LSaber36, line 104: So this can be improved. Since I showed you this code this is my bad, but here's why. Here we always check 4 if-statements. Remember, if statements are considered expensive. We can remove this if we just OR all of the conditions together. What I'm looking for is a structure similar to what is on line 118 of uart.ino.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -137,22 +151,39 @@ bool moveStraight(int motor[], int startCol, int startRow, int endCol, int endRo
...
+  pieceSpacesX = fabs(endCol - startCol);
   dirX = (endCol > startCol) ? RIGHT : LEFT;
 
-  spacesY = abs(endRow - startRow);
+  pieceSpacesY = fabs(endRow - startRow);
   dirY = (endRow > startRow) ? UP : DOWN;
 
-  numStepsX = spacesX * stepsPerSpace;
-  numStepsY = spacesY * stepsPerSpace;
+  numStepsX = 2 * (int)floor(pieceSpacesX) * stepsPerUnitSpace
```
LSaber36, line 179: You need a semicolon on this and the next line

PR #119, opened 2022-01-27
Update movement functions to work with ints and new movement scheme

adenm-10:


firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -83,32 +83,26 @@ void home()
 // Moves the magnet from the "start" point to the "end" point
 // This can only move in straight lines
 // Returns a boolean indicating success/error
-bool moveStraight(int motor[], float startCol, float startRow, float endCol, float endRow)
+bool moveStraight(int motor[], int startCol, int startRow, int endCol, int endRow)
 {
   // A specific motor is passed to this function since we are only moving one here
```
LSaber36, line 88: I only just now realized this, but this should be between lines 84 and 85

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -83,32 +83,26 @@ void home()
...
-    pieceSpaces = fabs(endCol - startCol);
+    unitSpaces = abs(endCol - startCol);
     dir = (endCol > startCol) ? RIGHT : LEFT;
     setScale(xMotor, WHOLE_STEPS);
   }
   else if (endCol == startCol)
   {
     // Y movement
-    pieceSpaces = fabs(endRow - startRow);
+    unitSpaces = abs(endRow - startRow);
```
LSaber36, line 105: On line 147 you have a comment about abs and how it's used, but not here. If you can make sure that "like" code blocks have the same comments, assuming they're applicable, please do so. It increases readability and consistency.

nashirj (approved):

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -63,6 +63,12 @@ int currentX, currentY;
 #define MS2_PIN 3
 #define ENDSTOP_PIN 4
 
+// Return definitions
```
LSaber36, line 66: Error codes

nashirj: For consistency's sake, what do you think of "Movement function status codes" rather than "Error codes" on and making this an enum instead of preprocessors (similar to how Chase set his stuff up, see [this](https://github.com/ucfai/knightros-gambit/blob/dec0e95f9112a3872012e5101d4dd0ad6e2f45e3/firmware/arduino/chessboard/circleFunction.ino#L5) or [this](https://github.com/ucfai/knightros-gambit/blob/dec0e95f9112a3872012e5101d4dd0ad6e2f45e3/firmware/arduino/chessboard/uart.ino#L10)).

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -82,45 +82,36 @@ void home()
 
 // Moves the magnet from the "start" point to the "end" point
 // This can only move in straight lines
-// Returns a boolean indicating success/error
-bool moveStraight(int motor[], float startCol, float startRow, float endCol, float endRow)
+// A specific motor is passed to this function since we are only moving one here
+// Returns a code indicating error or success circumstances
```
nashirj, line 86: If you update to an enum, you can have a more descriptive message about the return value like:
```c
// Returns a `MovementStatus` code, `0` if successful, nonzero
// for all errors (see `MovementStatus` in chessboard.ino).
```

firmware/arduino/chessboard/chessboard.ino
nashirj: PS: I don't think it matters in this case since we have multiple codes for error, but just a note that `0` is typically used for `false`/failure.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -131,49 +122,39 @@ bool moveStraight(int motor[], float startCol, float startRow, float endCol, flo
   // Rotate motor some number of steps
   for (i = 0; i < numSteps; i++) 
   {
-    if (digitalRead(X_AXIS_ENDSTOP_SWITCH) == HIGH  ||  
-        digitalRead(Y_AXIS_ENDSTOP_SWITCH) == HIGH)
-      return false;
+    if (digitalRead(X_AXIS_ENDSTOP_SWITCH) == HIGH)
+      return HIT_X_ENDSTOP;  
+    else if (digitalRead(Y_AXIS_ENDSTOP_SWITCH) == HIGH)
```
nashirj, line 127: This can just be an `if` statement.

If the first if statement is true, then you will return, (so the else if does nothing). However, if the first if statement is false, you'll check the next condition anyway (so the else if does nothing).

The [linter](https://pylint.org) I use for checking my python code always points this out when I do the same thing lol.

TL;DR, it isn't an error and it has same runtime, but I think for style you should use an `if` instead of an `else if` here.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -131,49 +122,39 @@ bool moveStraight(int motor[], float startCol, float startRow, float endCol, flo
...
 
-  return true;
+  return SUCCESS;
 }
 
 // Moves the magnet from the "start" point to the "end" point
 // This can move in diagonal lines of slopes: 1, 2, and 1/2
-// Returns a boolean indicating success/error
-bool moveDiagonal(float startCol, float startRow, float endCol, float endRow)
+// Returns a code indicating error and success circumstances
```
nashirj, line 140: Same as above about more meaningful comment if you change codes to an enum

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -197,14 +178,15 @@ bool moveDiagonal(float startCol, float startRow, float endCol, float endRow)
...
-    return false;
+    return INVALID_ARGS;
   }
 
   for (i = 0; i < numStepsX; i++)
   {
-    if (digitalRead(X_AXIS_ENDSTOP_SWITCH) == HIGH  ||  
-        digitalRead(Y_AXIS_ENDSTOP_SWITCH) == HIGH)
-      return false;
+    if (digitalRead(X_AXIS_ENDSTOP_SWITCH) == HIGH)
```
nashirj, line 186: Same comment as above with the `if` vs `else if`

nashirj (approved):
This code looks really good and much simpler than the float stuff! Nice work. I just gave a few small style suggestions but otherwise I'm fine with you merging this.

firmware/arduino/chessboard/motorDriver.ino
LSaber36: Even though it works both ways, having two if statements makes sense rather than the else statements. It has a similar scheme to checking for base cases in recursion

LSaber36:
I agree with all if what you said @nashirj

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -78,6 +72,15 @@ int currentX, currentY;
 // Number of different slope settings for each quarter circle
 #define NUM_SLOPES_PER_QUARTER_CIRCLE 9
 
+// Movement function status codes
+enum MovementStatus
+{
+  SUCCESS = '0',
+  HIT_X_ENDSTOP = '1',
+  HIT_Y_ENDSTOP = '2',
+  INVALID_ARGS = '3'
```
nashirj, line 81: These should be ints instead of chars, I know that's different from Chase's. sorry for not making that clear.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -83,7 +83,8 @@ void home()
 // Moves the magnet from the "start" point to the "end" point
 // This can only move in straight lines
 // A specific motor is passed to this function since we are only moving one here
-// Returns a code indicating error or success circumstances
+// Returns a 'MovementStatus' code, '0' if successful, varying nonzero values for various error codes
+// For all status codes, check 'MovementStatus in chessboard.ino
```
nashirj, line 87: Add closing quote for MovementStatus pls

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -137,7 +139,8 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
 
 // Moves the magnet from the "start" point to the "end" point
 // This can move in diagonal lines of slopes: 1, 2, and 1/2
-// Returns a code indicating error and success circumstances
+// Returns a 'MovementStatus' code, '0' if successful, varying nonzero values for various error codes
+// For all status codes, check 'MovementStatus in chessboard.ino
```
nashirj, line 143: closing quote

firmware/arduino/chessboard/chessboard.ino
nashirj:
```c
enum MovementStatus
{
   SUCCESS = 0,
   HIT_X_ENDSTOP = 1,
   HIT_Y_ENDSTOP = 2,
   INVALID_ARGS = 3
}
```

LSaber36: Also, since an enum can be declared anywhere (unlike preprocessors), this should be in motorDriver.ino

PR #128, opened 2022-01-31
Added doc-strings, fixed comments

fshcat:


chess-AI/train.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, lr, move_approximator, val_approximator=None):
         self.val_approximator = val_approximator
 
     def training_game(self):
+        """Runs a full game and returns policy and value output from approximators
```
nashirj, line 39:
```py
'''Run a full game, storing states, state values and policies for each state.

Returns a tuple of 3 lists where the ith element in each list corresponds to the same board state:
    1) fen strings representing board state
    2) state value predictions
    3) policy values for all legal moves
'''
```

nashirj (approved):
Looks great!

PR #129, opened 2022-01-31
Model checkpointing

JuddBE:
Saves the model to models.pt after every set of training and loads the previous model parameters to the network when train.py is run.


chess-AI/train.py
```diff
@@ -74,7 +73,7 @@ def training_game(self):
             # If the game is over, end the episode
             # TODO: Consider removing `board.can_claim_draw()` as it may be slow to check.
             # See https://python-chess.readthedocs.io/en/latest/core.html#chess.Board.can_claim_draw
-            if board.is_game_over() or board.can_claim_draw():
+            if board.is_game_over() or board.can_claim_fifty_moves() or board.can_claim_threefold_repetition() or board.is_seventyfive_moves():
```
fshcat, line 76: is this for sure necessary? everything else looks good

fshcat (approved):

chess-AI/train.py
JuddBE: can_claim_draw() performs the same function as can_claim_fifty_moves and can_claim_threefold_repetition(), so we could technically keep that, but it doesn't include the 75 move rule. If we want the game to stop as a draw from the 75 move rule we would need this part.

fshcat: isnt the 75 move rule a subset of the 50 move rule?

JuddBE: actually yeah 75 move rule is pointless if we check for 50 move rule, I'll get rid of it

chess-AI/train.py
```diff
@@ -155,22 +154,24 @@ def training_episode(self, nnet, games, epochs, batch_size):
 
                         # Reset gradients
                         opt.zero_grad()
-
                 print(losses)
+            save_model(nnet)
```
nashirj, line 158: Could optionally have some flag that specifies whether we want to save the new model.
```py
if saving_model:
    save_model(<new_parameters>)
```

nashirj: Maybe even optionally two flags, one `saving_model` and one `save_path`:
```py
if save_path or saving_model:
    save_model(model=nnet, save_path=save_path)
```
This way, if `save_path` is `None`, but we are saving the model, we save it with a name based on the most recently saved model (as described above), otherwise we save it with the specified path.

chess-AI/train.py
```diff
@@ -155,22 +154,24 @@ def training_episode(self, nnet, games, epochs, batch_size):
...
+
```
nashirj, line 161: Only one line between functions pls

chess-AI/checkpointing.py
```diff
@@ -0,0 +1,19 @@
+import torch
```
nashirj, line 1: Can you name this file `io.py` instead? That way, if you want to do any logging or other stuff later, it can all be contained in this file.

nashirj: It will also let us combine AI logging, game loop logging, etc in one file when we get further down the road.

JuddBE: > It will also let us combine AI logging, game loop logging, etc in one file when we get further down the road.

io is already an existing standard python library name so I'm not sure how I would import the functions, but I can change the name to be ai_io which is similar.

nashirj: Ah ok you can do that or leave as checkpointing.py whatever u prefer

ryPattillo (commented):
Looks good to merge

chess-AI/train.py
```diff
@@ -22,7 +22,7 @@ class Train:
         mcts: References the MCTS class
         policy_converter: References the PlayNetworkPolicyConverter class
     """
-    def __init__(self, lr, move_approximator, val_approximator=None):
+    def __init__(self, lr, move_approximator, save_path, val_approximator=None):
```
nashirj, line 25: Since specifying a save path is optional, let's have this parameter default to `None` e.g. `save_path=None`.

chess-AI/train.py
```diff
@@ -155,22 +158,46 @@ def training_episode(self, nnet, games, epochs, batch_size):
 
                         # Reset gradients
                         opt.zero_grad()
-
                 print(losses)
-
+            # Saves model to specified file, or a new file if not specified.
+            if self.save_path != None:
```
nashirj, line 163: `if self.save_path is not None`, see [style guide](https://google.github.io/styleguide/pyguide.html#2144-decision)

chess-AI/train.py
```diff
@@ -155,22 +158,46 @@ def training_episode(self, nnet, games, epochs, batch_size):
...
+            else:
+                for i in range(num_saved_models):
+                    if not(os.path.isfile(f'chess-AI/models-{i+1}.pt')):
+                        if overwrite_save and i != 0:
+                            save_model(nnet, f'chess-AI/models-{i}.pt')
+                            break
+                        save_model(nnet, f'chess-AI/models-{i+1}.pt')
+                        break
+                    if i == num_saved_models - 1:
+                        save_model(nnet, f'chess-AI/models-{num_saved_models}.pt')
```
nashirj, line 174: Would it be possible to have all this logic in the save function itself? I imagine this is something that might be used for stockfishtrain as well.

chess-AI/train.py
```diff
@@ -155,22 +158,46 @@ def training_episode(self, nnet, games, epochs, batch_size):
...
+    nnet.train()
+
+    if load_path != None:
+        nnet = load_model(nnet, load_path)
+    else:
+        for i in range(num_saved_models):
+            if not(os.path.isfile(f'chess-AI/models-{i+2}.pt')):
+                if i != 0:
+                    nnet = load_model(nnet, f'chess-AI/models-{i+1}.pt')
+                break
```
nashirj, line 192: Same comments as above.

chess-AI/train.py
```diff
@@ -100,7 +103,7 @@ def assign_rewards(self, board, length):
         print(board.outcome())
         return values
 
-    def training_episode(self, nnet, games, epochs, batch_size):
+    def training_episode(self, nnet, games, epochs, batch_size, num_saved_models, overwrite_save):
```
nashirj, line 106: I think we're getting to the point where we should define a `TrainOptions` class that we can pass a parameter to this function to make the function signature more clear and make it easier to add parameters later without modifying the function signature. Maybe we can work on that in a new PR after addressing all the other comments!
That way the function signature will always be something like
`def training_episode(self, nnet, train_ops):`
and we can have all the things currently being passed to the function as instance variables of the class.

chess-AI/train.py
nashirj, line 174: This is a cool idea! One downside is that different people may want a different number of saved models in training and they might replace someone else's saved model or something like that. I think having a way to check for the most recently saved model and creating filename based on that one would be good, what do you think @fshcat @ryPattillo

chess-AI/train.py
```diff
@@ -1,13 +1,13 @@
...
-import time
 
 import chess
+import os
 import torch
 import numpy as np
 from torch.utils.data import DataLoader, TensorDataset
 
 from mcts import Mcts
+from ai_io import save_model, load_model
```
nashirj, line 10: I realize that PEP 8 says it's discouraged to have underscores in module names. Can you change it back to `checkpointing` sorry pls don't hate me :laughing:

nashirj:
Sorry I know this is merged but you can address my comments in a new pr

Also can you please remove `models.pt` and add the `.gitignore` with the stuff we discussed yesterday? ty

PR #131, opened 2022-02-01
Implementing current position updating to primitives

adenm-10:
I was wondering if it would be better to make a setup function for numEightStepsX & Y, where they are declared and initialized to equal 8, then called upon in the primitives. Since the scales only include half or whole steps the division wouldn't have to include floats or anything.

### DIFF
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -98,6 +98,9 @@ void setup()
 
   stepsPerUnitSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
 
+  currentX = 0;
+  currentY = 0;
+
   // Initializes global 2d array `pulsesPerSlope` which is used to define circle paths
   // that are used in the `makeCircle()` function in circleFunction.ino
   calculatePulsesPerSlope();
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -98,6 +98,7 @@ void setup()
 
   stepsPerUnitSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
 
+  // Being initialized to 0 for safety
   currentX = 0;
   currentY = 0;
 
--- a/firmware/arduino/chessboard/motorDriver.ino
+++ b/firmware/arduino/chessboard/motorDriver.ino
@@ -65,8 +65,7 @@ void homeAxis(int motor[])
   // Moving motor towards home for rough estimate
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, WHOLE_STEPS);
-  currentMotorNumEighthSteps = 8;
-  currentMotorNumEighthSteps *= -1;
+  currentMotorNumEighthSteps = -8;
   while (digitalRead(motor[ENDSTOP_PIN]) == LOW)
   {
     // Moves motor
@@ -80,7 +79,7 @@ void homeAxis(int motor[])
 
   // Flips direction again to move motor away from home to prepare for fine-tuning
   digitalWrite(motor[DIR_PIN], RIGHT);
-  currentMotorNumEighthSteps *= -1;
+  currentMotorNumEighthSteps = 8;
   for (i = 0; i < HOME_CALIBRATION_OFFSET; i++)
   {
     digitalWrite(motor[STEP_PIN], LOW);
@@ -216,26 +215,24 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
     setScale(xMotor, WHOLE_STEPS);
     setScale(yMotor, WHOLE_STEPS);
     // Sets sign based off of direction of each motor
-    numEighthStepsX *= (dirX == RIGHT) ? 8 : -8;
-    numEighthStepsY *= (dirY == UP) ? 8 : -8;
+    numEighthStepsX = (dirX == RIGHT) ? 8 : -8;
+    numEighthStepsY = (dirY == UP) ? 8 : -8;
   }
   else if (numStepsY > numStepsX && (numStepsY / numStepsX) == 2)
   {
     setScale(xMotor, HALF_STEPS);
     setScale(yMotor, WHOLE_STEPS);
     // Sets sign based off of direction of each motor
-    numEighthStepsX *= (dirX == RIGHT) ? 4 : -4;
-    numEighthStepsY *= (dirY == UP) ? 8 : -8;
+    numEighthStepsX = (dirX == RIGHT) ? 4 : -4;
+    numEighthStepsY = (dirY == UP) ? 8 : -8;
   }
   else if (numStepsY < numStepsX && (numStepsX / numStepsY) == 2)
   {
     setScale(xMotor, WHOLE_STEPS);
     setScale(yMotor, HALF_STEPS);
-    numEighthStepsX = 8;
-    numEighthStepsY = 4;
     // Sets sign based off of direction of each motor
-    numEighthStepsX *= (dirX == RIGHT) ? 8 : -8;
-    numEighthStepsY *= (dirY == UP) ? 4 : -4;
+    numEighthStepsX = (dirX == RIGHT) ? 8 : -8;
+    numEighthStepsY = (dirY == UP) ? 4 : -4;
   }
   else
   {
--- a/firmware/arduino/chessboard/motorDriver.ino
+++ b/firmware/arduino/chessboard/motorDriver.ino
@@ -49,42 +49,55 @@ void disableMotors()
 // Drives the motor corresponding to "motor" to it's home position (position 0)
 void homeAxis(int motor[])
 {
+  int * currentMotorPos;
+  int currentMotorDir, currentMotorNumEighthSteps;
   int i;
+
+  currentMotorPos = (motor == xMotor) ? &currentX : &currentY;
 
   // Loop until endstop collision, then fine tune it
   // LEFT and DOWN have same value, as do RIGHT and UP. Use LEFT
   // and RIGHT arbitrarily while tuning end stop for both x and y axes.
 
+  // Sets scale and direction for motor and current position
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, WHOLE_STEPS);
+  currentMotorNumEighthSteps = 8;
+  currentMotorNumEighthSteps *= -1;
   while (digitalRead(motor[ENDSTOP_PIN]) == LOW)
   {
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+
+    // Motor is moved above, current position for whatever motor updated here
+    *currentMotorPos += currentMotorNumEighthSteps;
   }
 
   digitalWrite(motor[DIR_PIN], RIGHT);
+  currentMotorNumEighthSteps *= -1;
   for (i = 0; i < HOME_CALIBRATION_OFFSET; i++)
   {
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+   *currentMotorPos += currentMotorNumEighthSteps;
   }
 
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, EIGHTH_STEPS);
+  currentMotorNumEighthSteps = 1;
+  currentMotorNumEighthSteps *= -1;
   while (digitalRead(motor[ENDSTOP_PIN]) == LOW)
   {
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+    *currentMotorPos += currentMotorNumEighthSteps;
   }
 
-  // Initializing current position to 0 after homing function for safety,
-  // as all moves after will be based on the position given here
-  currentX = 0;
-  currentY = 0;
+  // Set the motor position to "home" position
+  *currentMotorPos = 0;
 }
 
 // Homes both axis
@@ -175,7 +188,7 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
   int numEighthStepsX, numEighthStepsY;
   int i;
 
-  // In following initializations and if statements, numEigthSteps direction and scale is set as well
+  // In following initializations and if statements, numEigthSteps scale is set
   // Abs ensures that numStepsX and numStepsY will be positive
   // to ensure proper for loop execution
   unitSpacesX = abs(endCol - startCol);
--- a/firmware/arduino/chessboard/motorDriver.ino
+++ b/firmware/arduino/chessboard/motorDriver.ino
@@ -8,7 +8,7 @@ enum MovementStatus
 };
 
 // Sets the scale of the motor driver corresponding to "motor"
-void setScale(int motor[], int scale) 
+void setScale(int motor[], int scale)
 {
   if (scale == WHOLE_STEPS)
   {
@@ -39,7 +39,7 @@ void enableMotors()
   digitalWrite(MOTOR_ENABLE, LOW);
 }
 
-void disableMotors() 
+void disableMotors()
 {
   digitalWrite(MOTOR_SLEEP, LOW);
   digitalWrite(MOTOR_RESET, LOW);
@@ -49,11 +49,11 @@ void disableMotors()
 // Drives the motor corresponding to "motor" to it's home position (position 0)
 void homeAxis(int motor[])
 {
-  int * currentMotorPos;
+  int *currentMotorPos;
   int currentMotorNumEighthSteps;
   int i;
 
-  //  Stores corresponding motor position based off of which motor is being homed,
+  // Stores corresponding motor position based off of which motor is being homed,
   // so that correct position can be incremented by function
   currentMotorPos = (motor == xMotor) ? &currentX : &currentY;
 
@@ -69,15 +69,16 @@ void homeAxis(int motor[])
   currentMotorNumEighthSteps *= -1;
   while (digitalRead(motor[ENDSTOP_PIN]) == LOW)
   {
+    // Moves motor
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
 
-    // Motor is moved above, current position for whatever motor updated here
+    // Updates current motor position
     *currentMotorPos += currentMotorNumEighthSteps;
   }
 
-  // Flipping direction again to move motor away from home to prepare for fine-tuning
+  // Flips direction again to move motor away from home to prepare for fine-tuning
   digitalWrite(motor[DIR_PIN], RIGHT);
   currentMotorNumEighthSteps *= -1;
   for (i = 0; i < HOME_CALIBRATION_OFFSET; i++)
@@ -85,14 +86,13 @@ void homeAxis(int motor[])
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
-   *currentMotorPos += currentMotorNumEighthSteps;
+    *currentMotorPos += currentMotorNumEighthSteps;
   }
 
-  // Moving motor towards home for fine-tuned home position
+  // Moves motor towards home for fine-tuned home position
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, EIGHTH_STEPS);
-  currentMotorNumEighthSteps = 1;
-  currentMotorNumEighthSteps *= -1;
+  currentMotorNumEighthSteps = -1;
   while (digitalRead(motor[ENDSTOP_PIN]) == LOW)
   {
     digitalWrite(motor[STEP_PIN], LOW);
@@ -101,7 +101,7 @@ void homeAxis(int motor[])
     *currentMotorPos += currentMotorNumEighthSteps;
   }
 
-  // Set the motor position to "home" position
+  // Sets the motor position to "home" position
   *currentMotorPos = 0;
 }
 
@@ -122,10 +122,10 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
   // How many steps per space
   int dir, numSteps, unitSpaces;
   int i;
-  int * currentMotorPos;
   int currentMotorNumEighthSteps;
+  int *currentMotorPos;
 
-  // Same as homeAxis(), sets the loop to only update a cingle motors position at a time
+  // Same as homeAxis(), sets the loop to only update a single motors position at a time
   // Direction is still determined seperately by if statements
   currentMotorPos = (motor == xMotor) ? &currentX : &currentY;
 
@@ -146,7 +146,7 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
     unitSpaces = abs(endRow - startRow);
     dir = (endRow > startRow) ? UP : DOWN;
     setScale(yMotor, WHOLE_STEPS);
-    //Sets motor and direction if Y movement
+    // Sets motor and direction if Y movement
     currentMotorNumEighthSteps = (dir == UP) ? 8 : -8;
   }
   else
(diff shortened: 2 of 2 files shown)


LSaber36:
I'm not sure what you mean here: "the division wouldn't have to include floats or anything."
But from what I'm gathering from the rest of your comment and code, it seems like you're seeing the repeated variables and are wondering if they can be reduced to one place to be a bit more streamlined.

LSaber36:
I have some ideas about how you could do this, but at the end of the day I think it makes more sense to keep those variables in their respective functions for now. The main reason is due to their scope; nothing outside of those functions needs to know about how many steps the motor is traveling with respect to the scale. Plus, if you think about the actual memory, you either have them always take up two bytes in memory (if global or something), or take up two bytes temporarily (if only used in the function). It's the same memory but one frees up more memory than the other. What are your thoughts @nashirj ?

adenm-10:
If it is better for the memory then that is perfect. All I was thinking about doing was making a function that would:
int numEighthStepsX = 8
int numEighthStepsY = 8,
as that declares and initializes once instead of twice. I didn't see any other opportunities to filter anything out though.
Also, I put at the end of the homing function to set currentX and currentY to 0, is that ok?

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -98,6 +98,9 @@ void setup()
 
   stepsPerUnitSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
 
+  currentX = 0;
```
LSaber36, line 101: Can you add a comment here explaining that you're initializing them to 0 here for safety?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -123,6 +127,11 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
 
   numSteps = unitSpaces * stepsPerUnitSpace;
 
+  numEighthStepsX = 8;
```
LSaber36, line 130: Can you make a comment explaining that you're setting the scales to 8 because this function always moves in whole steps?

LSaber36: Pls also comment that you set the sign based on the direction

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -155,6 +167,7 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
   int unitSpacesX, unitSpacesY;
   int dirX, dirY;
   int numStepsX, numStepsY;
+  int numEighthStepsX, numEighthStepsY;
   int i;
 
   // Abs ensures that numStepsX and numStepsY will be positive
```
LSaber36, line 173: Can you comment that, in the following blocks, you're setting the scale and assigning the number of eighth steps for X and Y?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -177,22 +190,31 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
...
     setScale(yMotor, HALF_STEPS);
+    numEighthStepsX = 8;
+    numEighthStepsY = 4;
   }
   else
   {
     return INVALID_ARGS;
   }
 
+  numEighthStepsX *= (dirX == RIGHT) ? 1 : -1;
```
LSaber36, line 215: Same as above

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -141,6 +150,9 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);  // 1 milliSecond
     digitalWrite(motor[STEP_PIN], HIGH);
+
+    currentX += numEighthStepsX;
```
LSaber36, line 154: Add a small comment saying "Updating current position" or something

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -206,6 +228,9 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
     delay(1);
     digitalWrite(xMotor[STEP_PIN], HIGH);
     digitalWrite(yMotor[STEP_PIN], HIGH);
+
+    currentX += numEighthStepsX;
```
LSaber36, line 232: Same as above

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -80,6 +80,9 @@ void homeAxis(int motor[])
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
   }
+
```
LSaber36, line 83: So this function is a bit special, since it passes in a specific motor. That being said, we need to differentiate between the motor arrays passed in, which we can do by comparing addresses. That comparison allows us to figure out what motor we're dealing with. We should also have a check for null in case it's passed through by accident, since we're dealing with addresses.

LSaber36: However, this only allows us to figure out what motor we're dealing with. We need to add to either of two possible variables, currentX and currentY. To do that, we'll use a pointer! If you need help lmk, cuz pointers are wack. But essentially, we'll make the temp variable hold the address of one of those variables, which then lets us deal with it arbitrarily. I have the code below.

LSaber36:
```c
void homeAxis(int motor[])
{
  // This is where you need to put these extra variables
  int * currentMotorPos;
  int currentMotorDir, currentMotorNumEighthSteps;
  int i;

  currentMotorPos = (motor == xMotor) ? &currentX : &currentY;

  // Do all of the motor control stuff
  // =====================
  // Here's one while loop for example
  // This one uses whole steps (8 1/8 steps)
  // and moves to the left
  currentMotorNumEighthSteps = 8;
  currentMotorNumEighthSteps *= -1;
  while (condition)
  {
    // Above this will be the code triggering the pulse to the motor driver
    *currentMotorPos += currentMotorNumEighthSteps;
  }

  // Set the motor position to "home" position
  *currentMotorPos = 0;  
}
```

LSaber36:
Also, on lines 23 and 30 of chessboard.ino, we can make those variables uint8_t instead of int. You'll need to change the function signatures that use these, which are pretty easy to identify if you change the data type and try to compile (it'll tell u where they don't match)

**Please do this in it's own commit**

firmware/arduino/chessboard/motorDriver.ino
LSaber36: should be numEighthSteps, not munEightSteps.

Remove: "moved at a time"

firmware/arduino/chessboard/motorDriver.ino
LSaber36: Don't need "for accuracy"

firmware/arduino/chessboard/motorDriver.ino
LSaber36: replace "is set as well" with "are set"

firmware/arduino/chessboard/motorDriver.ino
LSaber36: Same as above

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,42 +49,55 @@ void disableMotors()
...
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+
+    // Motor is moved above, current position for whatever motor updated here
+    *currentMotorPos += currentMotorNumEighthSteps;
   }
 
   digitalWrite(motor[DIR_PIN], RIGHT);
+  currentMotorNumEighthSteps *= -1;
```
LSaber36, line 78: This moves the motor away from home to fine tune, so this should be 1, not -1

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,42 +49,55 @@ void disableMotors()
...
   int i;
 
+  currentMotorPos = (motor == xMotor) ? &currentX : &currentY;
+
   // Loop until endstop collision, then fine tune it
   // LEFT and DOWN have same value, as do RIGHT and UP. Use LEFT
   // and RIGHT arbitrarily while tuning end stop for both x and y axes.
 
+  // Sets scale and direction for motor and current position
   digitalWrite(motor[DIR_PIN], LEFT);
```
LSaber36, line 63: While you're here, can you add this comment above this line:
// Moving motor towards home for rough estimate

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,42 +49,55 @@ void disableMotors()
...
+  currentMotorNumEighthSteps *= -1;
   for (i = 0; i < HOME_CALIBRATION_OFFSET; i++)
   {
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+   *currentMotorPos += currentMotorNumEighthSteps;
   }
 
   digitalWrite(motor[DIR_PIN], LEFT);
```
LSaber36, line 87: While you're here, can you add this comment above this line:
// Moving motor towards home for fine-tuned home position

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,42 +49,55 @@ void disableMotors()
 // Drives the motor corresponding to "motor" to it's home position (position 0)
 void homeAxis(int motor[])
 {
+  int * currentMotorPos;
+  int currentMotorDir, currentMotorNumEighthSteps;
   int i;
 
+  currentMotorPos = (motor == xMotor) ? &currentX : &currentY;
```
LSaber36, line 56: Please add a comment explaining that this is storing the corresponding motor position so we can increment the right one

firmware/arduino/chessboard/motorDriver.ino
LSaber36: *numEighthSteps

firmware/arduino/chessboard/motorDriver.ino
adenm-10: I set this to -1 because in the first calibration the direction was set to -1, so I was multiplying it by -1 again so that it would go right. As of that point currentMotorNumEighthSteps = -8, so if I left it currentX/Y would be going the wrong direction right?

firmware/arduino/chessboard/motorDriver.ino
adenm-10: Did I implement this correctly?

firmware/arduino/chessboard/motorDriver.ino
LSaber36: I'm glad you challenged my comment because you're totally right, can you add a comment explaining that? It can be with the other comment for organization.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -142,7 +147,7 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
 
   numSteps = unitSpaces * stepsPerUnitSpace;
 
-  // Setting munEightSteps moved at a time to 8 because function only moves in whole steps
+  // Setting numEightSteps to 8 because function only moves in whole steps
```
LSaber36, line 150: numEighthSteps

don't forget the 'h' pls

firmware/arduino/chessboard/motorDriver.ino
LSaber36: Functionally, it looks perfect! Nashir can double check before we fully merge but yeah it looks awesome. Thanks!

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,37 +49,60 @@ void disableMotors()
...
+  currentMotorNumEighthSteps *= -1;
   for (i = 0; i < HOME_CALIBRATION_OFFSET; i++)
   {
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+   *currentMotorPos += currentMotorNumEighthSteps;
   }
 
+  // Moving motor towards home for fine-tuned home position
```
nashirj, line 91: `Moves motor ...`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,37 +49,60 @@ void disableMotors()
...
   {
     digitalWrite(motor[STEP_PIN], LOW);
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+
+    // Motor is moved above, current position for whatever motor updated here
+    *currentMotorPos += currentMotorNumEighthSteps;
   }
 
+  // Flipping direction again to move motor away from home to prepare for fine-tuning
```
nashirj, line 80: In general, try to have comments that start with a verb `Does blah` instead of `Doing blah`. So in this case it would be `Flips direction ...`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,37 +49,60 @@ void disableMotors()
...
     delay(1);
     digitalWrite(motor[STEP_PIN], HIGH);
+   *currentMotorPos += currentMotorNumEighthSteps;
   }
 
+  // Moving motor towards home for fine-tuned home position
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, EIGHTH_STEPS);
+  currentMotorNumEighthSteps = 1;
+  currentMotorNumEighthSteps *= -1;
```
nashirj, line 95: You can replace these two lines with `currentMotorNumEighthSteps = -1`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,37 +49,60 @@ void disableMotors()
...
   // Loop until endstop collision, then fine tune it
   // LEFT and DOWN have same value, as do RIGHT and UP. Use LEFT
   // and RIGHT arbitrarily while tuning end stop for both x and y axes.
 
+  // Sets scale and direction for motor and current position
+  // Moving motor towards home for rough estimate
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, WHOLE_STEPS);
+  currentMotorNumEighthSteps = 8;
+  currentMotorNumEighthSteps *= -1;
```
nashirj, line 69: Replace these two lines with `currentMotorNumEighthSteps = -8;`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,37 +49,60 @@ void disableMotors()
...
+  //  Stores corresponding motor position based off of which motor is being homed,
+  // so that correct position can be incremented by function
+  currentMotorPos = (motor == xMotor) ? &currentX : &currentY;
+
   // Loop until endstop collision, then fine tune it
   // LEFT and DOWN have same value, as do RIGHT and UP. Use LEFT
   // and RIGHT arbitrarily while tuning end stop for both x and y axes.
 
+  // Sets scale and direction for motor and current position
+  // Moving motor towards home for rough estimate
```
nashirj, line 65: `Moves motor ...`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -123,6 +147,13 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
 
   numSteps = unitSpaces * stepsPerUnitSpace;
 
+  // Setting numEighthSteps to 8 because function only moves in whole steps
```
nashirj, line 150: `Sets numEighthSteps ...`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -123,6 +147,13 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
...
+  // Sign is set based off of direction of each motor
```
nashirj, line 151: `Sets sign based on direction of each motor`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -123,6 +147,13 @@ uint8_t moveStraight(int motor[], int startCol, int startRow, int endCol, int en
...
+  numEighthStepsX = 8;
+  numEighthStepsY = 8;
+  numEighthStepsX *= (dir == RIGHT) ? 1 : -1;
+  numEighthStepsY *= (dir == UP) ? 1 : -1;
```
nashirj, line 155: The way this is currently, you're using the same value of dir for the X and Y motion and updating both x and y position regardless of which direction you're moving.

`numEighthStepsX` and `numEighthStepsY` should be set inside the if statements. If x motion, do this:
```c
numEighthStepsX = (dir == RIGHT) ? 8 : -8;
numEighthStepsY = 0;
```
if y motion do this:
```c
numEighthStepsX = 0;
numEighthStepsY = (dir == UP) ? 8 : -8;
```

That way, you don't need to change the code below where you have
```c
// Updating current position per step
currentX += numEighthStepsX;
currentY += numEighthStepsY;
```
since if you're doing x motion, you're adding 0 to y and vice versa.

Alternatively, you can do something similar to what you did in `homeAxis` and use a pointer to update only one of the current positions, and you would set the value of that pointer in your if statements. Do you have a preference @LSaber36 ?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -155,8 +190,10 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
   int unitSpacesX, unitSpacesY;
   int dirX, dirY;
   int numStepsX, numStepsY;
+  int numEighthStepsX, numEighthStepsY;
   int i;
 
+  // In following blocks, scale and numEighthSteps for X and Y are set
```
nashirj, line 196: `Sets scale and numEighthSteps for both X and Y`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -177,22 +214,32 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
...
   }
   else
   {
     return INVALID_ARGS;
   }
 
+  // Sign is set based off of direction of each motor
+  numEighthStepsX *= (dirX == RIGHT) ? 1 : -1;
+  numEighthStepsY *= (dirY == UP) ? 1 : -1;
+
```
nashirj, line 242: This can be moved into the if statements, so each assignment will instead be
```c
// Sign is set based off of direction of each motor
numEighthStepsX = (dirX == RIGHT) ? MAGNITUDEX : -MAGNITUDEX;
numEighthStepsY = (dirY == UP) ? MAGNITUDEY : -MAGNITUDEY;
```
where `MAGNITUDEX` and `MAGNITUDEY` are the numbers you're assigning in each if statement

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -206,6 +253,10 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
     delay(1);
     digitalWrite(xMotor[STEP_PIN], HIGH);
     digitalWrite(yMotor[STEP_PIN], HIGH);
+
+    // Updating current position per step
```
nashirj, line 257: `Updates current position for both x and y`

nashirj (changes requested):

firmware/arduino/chessboard/motorDriver.ino
LSaber36: For consistency, we should mimic what we have in homeAxis() and use a pointer to refer to either of the directions. I didn't even realize the redundancy there, thanks!

firmware/arduino/chessboard/motorDriver.ino
LSaber36: You don't think it's better to assign magnitude then direction here? What you're suggesting would add more if statements and increase memory, but keep runtime the same.

firmware/arduino/chessboard/motorDriver.ino
nashirj: I agree with Lucas, let us know if you need help getting that worked out

firmware/arduino/chessboard/motorDriver.ino
LSaber36: We're going to assign magnitude and direction instead of decomposing it to optimize the runtime slightly, so Nashir's stuff is perfect!

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -49,37 +49,60 @@ void disableMotors()
 // Drives the motor corresponding to "motor" to it's home position (position 0)
 void homeAxis(int motor[])
 {
+  int * currentMotorPos;
+  int currentMotorDir, currentMotorNumEighthSteps;
   int i;
 
+  //  Stores corresponding motor position based off of which motor is being homed,
```
nashirj, line 56: Sorry for the additional comment - there should be one space before the start of the comment here instead of two

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -69,30 +69,30 @@ void homeAxis(int motor[])
...
+    *currentMotorPos += currentMotorNumEighthSteps;
   }
 
-  // Moving motor towards home for fine-tuned home position
+  // Moves motor towards home for fine-tuned home position
   digitalWrite(motor[DIR_PIN], LEFT);
   setScale(motor, EIGHTH_STEPS);
-  currentMotorNumEighthSteps = 1;
-  currentMotorNumEighthSteps *= -1;
+  currentMotorNumEighthSteps = -1;
```
LSaber36, line 95: This is perfect!

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -69,30 +69,30 @@ void homeAxis(int motor[])
   currentMotorNumEighthSteps *= -1;
```
LSaber36, line 69: For currentMotorNumEighthSteps and the other variables that do the same thing, Nashir and I decided on just assigning values to them. In this case, we'll be just setting them equal to the correct value, avoiding the multiplication.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -69,30 +69,30 @@ void homeAxis(int motor[])
...
 
-    // Motor is moved above, current position for whatever motor updated here
+    // Updates current motor position
     *currentMotorPos += currentMotorNumEighthSteps;
   }
 
-  // Flipping direction again to move motor away from home to prepare for fine-tuning
+  // Flips direction again to move motor away from home to prepare for fine-tuning
   digitalWrite(motor[DIR_PIN], RIGHT);
   currentMotorNumEighthSteps *= -1;
```
LSaber36, line 83: Same as above, this should be setting equal to the expected value of 1 instead of essentially multiplying by -1 twice

firmware/arduino/chessboard/motorDriver.ino
LSaber36, line 95: Perfect! This is exactly how the other assignments should look.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -210,41 +210,42 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
...
-  
+
   if (numStepsX == numStepsY)
   {
     setScale(xMotor, WHOLE_STEPS);
     setScale(yMotor, WHOLE_STEPS);
-    numEighthStepsX = 8;
-    numEighthStepsY = 8;
+    // Sets sign based off of direction of each motor
+    numEighthStepsX *= (dirX == RIGHT) ? 8 : -8;
```
LSaber36, line 219: Same as above. You just need to remove that * in the *= operation.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -210,41 +210,42 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
...
+    numEighthStepsY *= (dirY == UP) ? 8 : -8;
   }
   else if (numStepsY > numStepsX && (numStepsY / numStepsX) == 2)
   {
     setScale(xMotor, HALF_STEPS);
     setScale(yMotor, WHOLE_STEPS);
-    numEighthStepsX = 4;
-    numEighthStepsY = 8;
+    // Sets sign based off of direction of each motor
+    numEighthStepsX *= (dirX == RIGHT) ? 4 : -4;
```
LSaber36, line 227: Same as above

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -210,41 +210,42 @@ uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
...
+    numEighthStepsY *= (dirY == UP) ? 8 : -8;
   }
   else if (numStepsY < numStepsX && (numStepsX / numStepsY) == 2)
   {
     setScale(xMotor, WHOLE_STEPS);
     setScale(yMotor, HALF_STEPS);
     numEighthStepsX = 8;
     numEighthStepsY = 4;
+    // Sets sign based off of direction of each motor
+    numEighthStepsX *= (dirX == RIGHT) ? 8 : -8;
```
LSaber36, line 237: Same as above, but you also need to remove those assignments above this, they're unnecessary.

PR #134, opened 2022-02-01
Cv dataset helper files

### LINKED ISSUE #112: Implement helper files needed for computer vision dataset collection
See the README.md introduced in #111. This file outlines a procedure for building a cv dataset. Two files need to be implemented to support this procedure:
1) `get_images_of_game.py`: This is a simple CLI type thing where video of the board is streaming from overhead and user presses a key on the keyboard to take a picture of the current board, which is then saved to disk (move `j` is saved as `j.png`).
2) `extract_labels_from_game.py`: This file (for each move in a game) uses the fen string to compute a 2d grid representation of the board, transforms and segments the image corresponding to that move, then saves that creates a label
Here is some pseudocode for 2)
```python
for each j, (image, fen) in enumerate(zip(game_i.images, game_i.fens)):
    create subdirectory "game_i/j"
    segmented_2d_arr = sqs_from_full_board_image(image)
    board_2d = get_2d_grid(fen)
    for i in range(8):
        for j in range(8):
            color_label, type_label = get_labels(board_2d[i][j])
            # i and j are used in creating appropriate filename
            save_img_and_labels(i, j, segmented_2d_arr[i][j], color_label, type_label)
            # Each cell yields a single image with path `game_i/j/<sq>` where sq is the cell on the chessboard
            # for that piece (e.g. `e4`). Two labels are generated for each `sq`
            # 0, 1, 2; these correspond to '.', 'w', 'b') and are used for square occupancy detection
            # 0, 1, 2, 3, 4; these correspond to 'p', 'q', 'b', 'n', 'r' and are used for piece classification
```

nashirj:
These files will be used to build a dataset to train a piece and type classifier. They are a prereq to #101 and #102 , though those will be addressed in separate PRs.

This closes #112

### DIFF
--- a/computer-vision/data/kg_dataset/README.md
+++ b/computer-vision/data/kg_dataset/README.md
@@ -7,10 +7,10 @@
 - Save pgn from the game to the `game_i` subdirectory
 - Create a `full_game` subdirectory (`game_i/full_game`)
 - Use `pgn-extract` to extract one fen for each PGN into the `full_game` subdirectory
-	- Label these fens `j.fen` where `j` corresponds to the move number of `game_i` when this fen is the board state (Note: `j` starts at 0, and `move_j` is just the starting board state)
+	- Label these fens `j.fen` where `j` corresponds to the move number of `game_i` when this fen is the board state (Note: `j` starts at 0, and move `j` is just the starting board state)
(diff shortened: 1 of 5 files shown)


ryPattillo (commented):
Ready to Merge

PR #136, opened 2022-02-03
adds gitignore for saved models

JuddBE:


nashirj:
Looks good, but can you add a comment saying `Ignore all saved model files with exceptions below` and then below the wildcard exclude have a comment saying `Saved models to keep. Note: when adding models here, please use descriptive names)` for every exception.
So the file will look like:
```gitignore
# Ignore all saved model files with exceptions below
*.pt

# Saved models to keep. Note: when adding models here, add a description indicating why model is kept
# Our first model!
!models.pt
```

Issue #31, opened 2021-11-15
Store past board states in game loop


nashirj:
Need to store 8 past board states as 2d grid (needed for RL to check repetition)
Need to store 1 past board state as image (needed for CV to compute pixel diff)

We could compute the board state as 2d grid at each step and save this within a [deque](https://www.geeksforgeeks.org/deque-in-python/) in the Board class

PR #139, opened 2022-02-03
Track previous states in game loop and minimize repeated creation of 2d grid.

### LINKED ISSUE #104: Minimize repeated calls to get_2d_board
`util.get_2d_board` creates and returns a 2d array. Store the 2d array at the start of every "new" move, and then we can simply pass this 2d array around. This can be addressed in the same PR as issue #31

nashirj:
Closes #104 and #31


JuddBE (approved):
Looks good!

PR #146, opened 2022-02-08
Update primitive signatures for current location tracking

adenm-10:


adenm-10:
I was wondering if there was a way to only do one conversion from currentX/Y to startRow/Col using pointers like we did to know which motor to track for *currentMotorPosition?

LSaber36:
Since they need to be unique, we don't want to use a pointer. We used a pointer before since the motor variable wasn't unique (meaning a specific motor)

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -43,6 +43,7 @@ uint8_t yMotor[5] = {Y_MOTOR_STEP, Y_MOTOR_DIR, Y_MOTOR_MS1, Y_MOTOR_MS2, Y_AXIS
 #define STEPS_PER_MILLIMETER 5
 #define HOME_CALIBRATION_OFFSET 100
 float stepsPerUnitSpace;
+float eighthStepsPerUnitSpace;
```
LSaber36, line 46: We don't need this, see my below comments

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -96,7 +97,9 @@ void setup()
   pinMode(Y_AXIS_ENDSTOP_SWITCH, INPUT);
   pinMode(CHESS_TIMER_BUTTON, INPUT);
 
+  // Defines different units of board being used
   stepsPerUnitSpace = MILLIMETERS_PER_UNITSPACE * STEPS_PER_MILLIMETER;
+  eighthStepsPerUnitSpace = stepsPerUnitSpace * 8;
```
LSaber36, line 102: Since we only use eighthStepsPerUnitSpace in a small number of places, we can avoid using a variable for this and just do the assignment directly., like below:
```
  startCol = currentX / (stepsPerUnitSpace * 8);
  startRow = currentY / (stepsPerUnitSpace * 8);
```

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -129,13 +129,18 @@ void home()
...
+uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
 {
   // How many steps per space
   int dir, numSteps, unitSpaces;
   int i;
   int eighthStepsPerPulse;
   int *currentMotorPos;
+  float startCol, startRow;
+
+  // Converts current position to be in terms of unit spaces
```
LSaber36, line 141: Just make sure to add to your comment explaining what the (stepsPerUnitSpace * 8) means

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -198,13 +203,18 @@ uint8_t moveStraight(uint8_t motor[], int startCol, int startRow, int endCol, in
...
+uint8_t moveDiagonal(int endCol, int endRow)
 {
   int unitSpacesX, unitSpacesY;
   int dirX, dirY;
   int numStepsX, numStepsY;
   int eighthStepsPerPulseX, eighthStepsPerPulseY;
   int i;
+  float startRow, startCol;
+
+  // Converts current position to be in terms of unit spaces
```
LSaber36, line 215: Same as above

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -96,7 +97,9 @@ void setup()
   pinMode(Y_AXIS_ENDSTOP_SWITCH, INPUT);
   pinMode(CHESS_TIMER_BUTTON, INPUT);
 
+  // Defines different units of board being used
```
LSaber36, line 100: Having a comment here is great!

LSaber36 (changes requested):

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -97,9 +96,8 @@ void setup()
   pinMode(Y_AXIS_ENDSTOP_SWITCH, INPUT);
   pinMode(CHESS_TIMER_BUTTON, INPUT);
 
-  // Defines different units of board being used
+  // Defines units of board being used
```
LSaber36, line 99: This is just my preference, but I like to avoid using the format "A of B" and just use "B's A"
In this case, it would be "the board's units" instead of "units of board"

LSaber36 (changes requested):

LSaber36 (approved):

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -129,13 +129,18 @@ void home()
...
 // For all status codes, check 'MovementStatus' in chessboard.ino
-uint8_t moveStraight(uint8_t motor[], int startCol, int startRow, int endCol, int endRow)
+uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
 {
   // How many steps per space
   int dir, numSteps, unitSpaces;
   int i;
   int eighthStepsPerPulse;
   int *currentMotorPos;
+  float startCol, startRow;
```
nashirj, line 139: Can you please make these `int`?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -198,13 +203,18 @@ uint8_t moveStraight(uint8_t motor[], int startCol, int startRow, int endCol, in
...
 // For all status codes, check 'MovementStatus' in chessboard.ino
-uint8_t moveDiagonal(int startCol, int startRow, int endCol, int endRow)
+uint8_t moveDiagonal(int endCol, int endRow)
 {
   int unitSpacesX, unitSpacesY;
   int dirX, dirY;
   int numStepsX, numStepsY;
   int eighthStepsPerPulseX, eighthStepsPerPulseY;
   int i;
+  float startRow, startCol;
```
nashirj, line 213: Can you please make these `int`?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -129,13 +129,18 @@ void home()
...
   // How many steps per space
   int dir, numSteps, unitSpaces;
   int i;
   int eighthStepsPerPulse;
   int *currentMotorPos;
+  float startCol, startRow;
+
+  // Converts current position to be in terms of unit spaces instead of eighth steps
+  startCol = currentX / (stepsPerUnitSpace * 8);
+  startRow = currentY / (stepsPerUnitSpace * 8);
```
nashirj, line 143: Here and below, it would be good to have a check to make sure that `currentX` and `currentY` are aligned with a unit space so that we know if something funky is going on with the positioning of the electromagnet. Right now, if they are not aligned, we have no way of being able to tell. @LSaber36 said that he prefers to address it later, here is a link to an issue regarding that change: #147

nashirj (commented):
Can you add a comment to line 43 please:
```c
#define STEPS_PER_MILLIMETER 5  // Whole steps per millimeter
```

firmware/arduino/chessboard/motorDriver.ino
LSaber36: After considering it, I agree with Nashir. Though, it would be a good idea to make this in a different issue, since it will involve either adding an error code, which means enums, or error correction, which involves some movement code.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -136,7 +136,12 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
   int i;
   int eighthStepsPerPulse;
   int *currentMotorPos;
-  float startCol, startRow;
+  int startCol, startRow;
+
+  // Checks if the EM is aligned properly
+  if((currentX % stepsPerUnitSpace) || (currentY % stepsPerUnitSpace)) {
+    return INVALID_ARGS;
```
LSaber36, line 143: Can you make a new error code called INVALID_ALIGNMENT and return that instead?

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -140,7 +141,7 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
 
   // Checks if the EM is aligned properly
   if((currentX % stepsPerUnitSpace) || (currentY % stepsPerUnitSpace)) {
```
LSaber36, line 143: if statements should be written as:
if (condition)
{

}

You need to put the one bracket on a newline and add a space after "if"

PR #154, opened 2022-02-11
Test what edge detection algo is best

ColonelToad:


nashirj (commented):
Could you please make a Jupyter notebook showing your results with these changes instead? I think that would be really helpful so that we can know how these different algorithms are performing when we use them to detect corners.

nashirj:
I'm going to close this since we aren't updating the files in this PR yet.

PR #155, opened 2022-02-11
Remove train class

fshcat:
Got rid of the class and moved parameters to functions

### DIFF
--- a/chess-AI/train.py
+++ b/chess-AI/train.py
@@ -15,236 +15,200 @@ from output_representation import PlayNetworkPolicyConverter
 from state_representation import get_cnn_input
 from stockfish_train import StockfishTrain
 
-class Train:
-    """The main training class, will train using stockfish and mcts
-    
-    1. Will create a dataset or load a dataset of (fen_strings,move_probs,values)
-    generated from Stockfish
-    2. Will then train using the model
-    3. Then will generate training examples by self play using the MCTS
-    4. Will train the model on those examples
 
-    Attributes:
-        policy_converter: References the PlayNetworkPolicyConverter class
-        learning_rate: The learning rate for the model
-        move_approximator: The lambda function that gets move probs
-        value_approximator: The lambda function that gets the value
-        save_path: The location where to save the model
-        device: The device used for training (either CPU or CUDA)
+def training_game(val_approximator, move_approximator):
+    """Run a full game, storing fen_strings, policies, and values
+
+    Returns a tuple of 3 lists where the ith element in each list corresponds to the board state:
+        1) fen strings representing board state
+        2) policy values for all legal moves
+        3) state value predictions
     """
 
-    def __init__(self, learning_rate, move_approximator, save_path, device, val_approximator=None):
+    # Stores probability distributions and values from the approximators
+    all_move_probs = []
+    all_state_values = []
 
-        # Policy Converter to convert output of the model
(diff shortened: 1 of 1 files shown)


JuddBE:
Looks good to merge. After this is merged I have the code ready for the model saving reorganization which I can add right after.

chess-AI/train.py
```diff
@@ -15,231 +15,209 @@
...
+    Attributes:
+        dataset: the dataset to use for training
+        dashboard: the streamlit dashboard (this will eventually be removed)
+        nnet: the neural network
+        learning_rate: the learning rate for the optimizer
+        epochs: number of epochs
+        batch_size: the batch size for SGD
+        device: the device being used to train (either CPU or GPU)
+        save_path: path for model checkpointing
+        num_saved_models: number of models to store
```
nashirj, line 133: I know this is how it was before this PR, but can you please order this according to the order of the passed parameters and remove `dashboard` from the docstring

chess-AI/train.py
```diff
@@ -15,231 +15,209 @@
...
+    average_pol_loss = []
+    average_val_loss = []
+
+    # Define loss functions (cross entropy and mean squared error)
+    ce_loss_fn = torch.nn.CrossEntropyLoss()
+    mse_loss_fn = torch.nn.MSELoss()
+
+    # Create optimizer for updating parameters during training
+    # TODO: Consider using other optimizers, such as Adam
+    opt = torch.optim.SGD(nnet.parameters(), lr=learning_rate, weight_decay=0.001, momentum=0.9)
```
nashirj, line 146: Again, I know this is how it was before the PR, but should we pass `weight_decay` and `momentum` as parameters?

chess-AI/train.py
```diff
@@ -249,19 +227,18 @@ def main():
...
-    stocktrain_games, stocktrain_epochs = 1000,10
-    mcts_games, mcts_epochs = 100,100
-    batch_size, learning_rate = 8,0.1
+    stocktrain_games, stocktrain_epochs = 1000, 10
+    mcts_games, mcts_epochs = 100, 100
+    batch_size, learning_rate = 8, 0.1
 
     # Gets stockfish training object, and sets parameters (elo,depth)
-    stockfish_path = "C:/Users/juddb/Stockfish/stockfish_14.1_win_x64_avx2/stockfish_14.1_win_x64_avx2.exe"
+    stockfish_path = r"C:\Users\jhoyo\Downloads\stockfish_14.1_win_x64_avx2.exe"
```
nashirj, line 235: Doesn't need to be part of this PR, but can all the Windows people @JuddBE @fshcat move the stockfish binary to "../../chess-engine/stockfish_14.1_win_x64_avx2.exe" so that the path doesn't change between PRs? As part of that change, we can update this initialization to check the system and set the stockfish path accordingly (like we do in [`create_stockfish_wrapper`](https://github.com/ucfai/knightros-gambit/blob/master/firmware/raspi/util.py#L38)). I can add that in a new PR, just wanted to let you guys know so that you can move the binary to the right place.

nashirj (approved):

chess-AI/train.py
nashirj: I'll do the initialization thing after this PR is merged, but for now can you just update this line to reflect the path I put in the comment above?

PR #156, opened 2022-02-12
Add extreme endstops

adenm-10:


nashirj (approved):
Only comment I have is that since the origin (0, 0) of the board is at the bottom left of the board, I'm not sure about using POS and NEG. Maybe using X_AXIS_ZERO_ENDSTOP and X_AXIS_MAX_ENDSTOP (same for Y) is more clear? I think that naming reflects the configuration of the end stops on the board a little more clearly. Otherwise this looks good to go, awesome work Aden!

PR #158, opened 2022-02-13
Create stockfish wrapper to allow system agnostic creation of Stockfish object in stockfish_train

nashirj:


chess-AI/stockfish_train.py
```diff
@@ -15,31 +19,22 @@ class StockfishTrain:
...
         """Sets the elo and depth for stockfish using the dashboard
 
         Attributes:
           dashboard: reference to the streamlit dashboard (this is temporary)
         """
-        
+
         # Set the elo and depth of the stockfish object
         self.stockfish.set_elo_rating(1000)
         self.stockfish.set_depth(3)
```
fshcat, line 34: idk if this should go in a separate pr but elo and depth should be constructor arguments imo, i dont think set_params should exist

fshcat (approved):

chess-AI/stockfish_train.py
nashirj: Agree, I'll do it in this one

PR #162, opened 2022-02-15
io reorganized

JuddBE:


chess-AI/ai_io.py
```diff
@@ -1,14 +1,34 @@
 import torch
+import os
```
nashirj, line 2: Import order should be:
```py
import os

import torch
```
since os is builtin and torch is third party

chess-AI/ai_io.py
```diff
@@ -1,14 +1,34 @@
...
 
-def save_model(nnet, save_path):
+def save_model(nnet, save_path, num_saved_models, overwrite):
```
nashirj, line 4: Please change all paths to `./models/...`

nashirj:
Additional comment: if `models` and `datasets` are not folders already, the code should make those folders.

chess-AI/ai_io.py
```diff
@@ -1,14 +1,34 @@
...
-    nnet.load_state_dict(torch.load(f'{save_path}'))
+    if model_path is not None:
+        nnet.load_state_dict(torch.load(model_path))
+    else:
+        for i in range(num_saved_models):
+            if not (os.path.isfile(f'chess-AI/models-{i + 2}.pt')):
+                if i != 0:
+                    nnet.load_state_dict(torch.load(f'chess-AI/models-{i + 1}.pt'))
+                break
     return nnet
```
fshcat, line 32: does this function need to return? pretty sure nnet is passed by reference so it should modify it directly

PR #163, opened 2022-02-16
Add arg parser to game

nashirj:
Add `argparser` to game loop based on Lucas's code:
```py
import argparse

def sendMessageToMicrocontroller():
  print('Sending message to microcontroller')

def sendMessageToChessEngine():
  print('Sending message to chess engine')

def parseMessageFromCli():
  print('Parsing message from CLI:')
  print('''
  DEPENDENCY:
  This message will vary depending on the microcontroller flag''')


def parseMessageFromTestFiles():
  print('''
  TODO: Make it so that if you have the -t or --test flag, 
        you MUST also have a file name.''')

  print('''
  TODO: Use sub-parsers to make a command line argument 
        dependency for an additional -a or --all command''')
  
  print('''
  DEPENDENCY:
  This message will vary depending on the microcontroller flag''')

def parseMessageFromHardware():
  print('Parsing message from hardware')

def main():
  parser = argparse.ArgumentParser(description='These allow us to select how we want the game to be played',
                                   epilog='''The default is to run the chess engine using inputs from player moves
                                             detected by computer vision''')

  parser.add_argument('-d', '--debug', 
                      dest='debug',
                      action='store_const',
                      const=True,
                      default=False, 
                      help='makes the chess engine parse moves from the cli')

  parser.add_argument('-t', '--test',
                      dest='test',
                      action='store_const',
                      const=True, 
                      default=False, 
                      help='makes the chess engine parse moves from a test files')

  parser.add_argument('-m', '--microcontroller', 
                      dest='microcontroller',
                      action='store_const',
                      const=True,
                      default=False, 
                      help='sends commands over UART directly, instead of through the chess engine')

  args = parser.parse_args()
  
  # Print newline for spacing
  print()

  # Decides the source of the message
  if (args.debug):
    parseMessageFromCli()
  elif (args.test):
    parseMessageFromTestFiles()
  else:
    parseMessageFromHardware()

  # Print newline for spacing
  print()

  # Decides the destination for the messages
  if (args.microcontroller):
    sendMessageToMicrocontroller()
  else:
    sendMessageToChessEngine()
  
  # Print newline for spacing
  print()

if __name__ == '__main__':
  main()
```


firmware/raspi/game.py
```diff
@@ -68,15 +70,63 @@ def player_wants_rematch():
...
+# TODO: abstract main program logic in here
+def run_user_interaction_loop():
+    pass
+
+# TODO: abstract test file logic in here
+def run_test_file():
+    pass
+
+def init_parameters():
+    args = parse_args()
```
nashirj, line 115: Move `parse_args` to be right above this function

PR #169, opened 2022-02-19
Demote current position calculation out of loops

adenm-10:
…duce computation

### DIFF
--- a/firmware/arduino/chessboard/motorDriver.ino
+++ b/firmware/arduino/chessboard/motorDriver.ino
@@ -239,10 +239,10 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
     digitalWrite(motor[STEP_PIN], HIGH);
   }
 
-    // Updates current position of relevant motor
-    // Not incremented inside of loop to save runtime, and isn't needed as endstop collision
-    // is only exception to end position result, which will be dealt with by calling alignAxis
-    *currentMotorPos += (eighthStepsPerPulse * numSteps);
+  // Updates current position of relevant motor
+  // Not incremented inside of loop to save runtime and unneeded computation
+  // If motor collides with endstop, alignAxis is triggered and fixes motor position
+  *currentMotorPos += (eighthStepsPerPulse * numSteps);
 
   return SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


adenm-10:
For the calculations inside the loops of alignAxis, should I just get rid of them all together since we're setting position equal to 0 anyway or create a variable that keeps track of how many iterations of each loop occur and implement current position assignments after each loop? Also I feel like there should be more to my comments but that's the only benefit of moving the calculation I could think of.

LSaber36:
For now, you can just assign the position at the end of alignAxis. Nothing can interrupt the code for that, except for ISRs. Even so, an ISR would resume the code in that function if interrupted, so the flow would be the same.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -242,11 +237,13 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
...
     digitalWrite(motor[STEP_PIN], HIGH);
-
-    // Updating current position per step
-    *currentMotorPos += eighthStepsPerPulse;
   }
 
+    // Updates current position of relevant motor
+    // Not incremented inside of loop to save runtime, and isn't needed as endstop collision
+    // is only exception to end position result, which will be dealt with by calling alignAxis
+    *currentMotorPos += (eighthStepsPerPulse * numSteps);
```
nashirj, line 245: Indentation is off here, need to decrease by two spaces. Also line 243 and 244 are a little confusing can u pls change it to one statement per line, something like:
```c
// Updated position is calculated outside of loop to reduce computation/runtime
// If motor collides with endstop, `alignAxis` is called to fix current position
```

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -338,11 +335,13 @@ uint8_t moveDiagonal(int endCol, int endRow)
...
-    // Updates current position for both x and y
-    currPositionX += eighthStepsPerPulseX;
-    currPositionY += eighthStepsPerPulseY;
   }
 
+    // Updates current position for both X and Y motors
+    // Not incremented inside of loop to save runtime, and isn't needed as endstop collision
+    // is only exception to end position result, which will be dealt with by calling alignAxis
+    currPositionX += (eighthStepsPerPulseX * numStepsX);
+    currPositionY += (eighthStepsPerPulseY * numStepsX);
```
nashirj, line 344: Fix indentation here as well, and update comment as above.

nashirj (commented):
Overall looks really good, just left you two comments.

PR #182, opened 2022-02-25
Made training flags, split stock dataset creation

### LINKED ISSUE #172: Update train_on_stockfish in train.py to not create/load dataset
The aim of this change is to split up the functionality of train.py into 3 distinct steps: 1) create dataset, 2) train on stockfish dataset, 3) train on mcts. Then, we can specify command line flags to perform any combination of these 3 steps.

As part of this need to update `train_on_stockfish` to
- not create/load dataset
- have a dataset parameter

We also need to add flags to train.py that indicate whether to run any combination of the three steps above

fshcat:
Resolves #172

### DIFF
--- a/chess-AI/train.py
+++ b/chess-AI/train.py
@@ -261,58 +261,65 @@ def main():
     device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
     nnet = PlayNetwork().to(device=device)
 
-
     nnet, dataset_path, stockfish_options, mcts_options, start_train, show_dash, make_dataset_flag, stockfish_train_flag, mcts_train_flag= init_params(nnet, device)
 
-    # Makes parent directories if necessary
-    make_dir(dataset_path)
-
-    if start_train:
-        if make_dataset_flag:
-            msg = "Dataset Creation Has Begun"
-            if show_dash:
-                Dashboard.info_message("success", msg)
-            else:
-                print(msg)
-            dataset = create_stockfish_dataset(stockfish_options, show_dash)
-            torch.save(dataset, dataset_path)
-
-            msg = "Dataset Creation completed"
-            if show_dash:
-                Dashboard.info_message("success", msg)
-            else:
-                print(msg)
+    # Need to check for dataset
+    if not dataset_path:
+        msg = "No dataset path specified"
+        if show_dash:
+            Dashboard.info_message("error", msg)
         else:
-            dataset = load_dataset(dataset_path, show_dash)
+            print(msg)
+    else:
+        # Makes parent directories if necessary
+        make_dir(dataset_path)    
 
-        if stockfish_train_flag:
-            msg = "Stockfish Training Has Begun"
-            if show_dash:
-                Dashboard.info_message("success", msg)
-            else:
-                print(msg)
-            train_on_dataset(dataset, nnet, stockfish_options)
+        if start_train:
+            if make_dataset_flag:
+                msg = "Dataset Creation Has Begun"
+                if show_dash:
+                    Dashboard.info_message("success", msg)
+                else:
+                    print(msg)
+                dataset = create_stockfish_dataset(stockfish_options, show_dash)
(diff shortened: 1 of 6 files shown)


chess-AI/train.py
```diff
@@ -260,35 +249,68 @@ def main():
...
-    nnet, dataset_path, stockfish_options, mcts_options, start_train, show_dash = init_params(nnet, device)
-
-    if start_train:
-        msg = "Stockfish Training Has Begun"
-        if show_dash:
-            Dashboard.info_message("success", msg)
-        else:
-            print(msg)
-        train_on_stockfish(nnet, dataset_path, stockfish_options, show_dash)
+    make_dataset_flag, stockfish_train_flag, mcts_train_flag = True, True, True
```
nashirj, line 252: These should be returned from `init_params` and added as flags to the argparser. If we don't add them to the dashboard in this PR (which I would prefer we do), let's default to setting them true if `args.dashboard` is `True`. This would affect `train.sh` so we could update that as well to handle these additional CLI arguments.

chess-AI/train.py
```diff
@@ -215,29 +219,14 @@ def train_on_dataset(dataset, nnet, options, show_dash=False):
     save_model(nnet, options.save_path, options.num_saved_models, options.overwrite)
 
 
-def train_on_stockfish(nnet, dataset_path, sf_opt, show_dash=False):
+def create_stockfish_dataset(sf_opt, show_dash):
```
nashirj, line 222: pls add docstring

chess-AI/train.py
```diff
@@ -215,29 +219,14 @@ def train_on_dataset(dataset, nnet, options, show_dash=False):
...
-        # TODO: Dataset should be given a more descriptive name, this is just temporary
-        # TODO: This should be in the ai_io file
-        torch.save(dataset, './datasets/stockfish_data.pt')
-
-    # Train using the stockfish dataset
-    train_on_dataset(dataset, nnet, sf_opt)
+    return create_dataset(sf_opt.games, stocktrain_moves, stocktrain_value_approximator, show_dash)
 
 
 def train_on_mcts(nnet, mcts_opt, show_dash=False):
```
nashirj, line 232: pls add docstring

chess-AI/train.py
```diff
@@ -260,35 +249,68 @@ def main():
...
-            print(msg)
+            assert path.exists(dataset_path), "Dataset not found at path provided."
+
+            msg = "Dataset retrieved."
+            if show_dash:
+                Dashboard.info_message("success", msg)
+            else:
+                print(msg)
+
+            dataset = torch.load(dataset_path)
```
nashirj, line 284: pls move this stuff to `ai_io` if it makes sense to do so

chess-AI/train.py
```diff
@@ -260,35 +249,68 @@ def main():
...
-        # Train network using MCTS
-        msg = "MCTS Training has begun"
-        if show_dash:
-            Dashboard.info_message("success", msg)
-        else:
-            print(msg)
-        train_on_mcts(nnet, mcts_options, show_dash)
+    # Makes parent directories if necessary
+    if not path.exists(path.dirname(dataset_path)):
+        makedirs(path.dirname(dataset_path))
```
nashirj, line 258: pls move this to `ai_io` also

chess-AI/train.py
```diff
@@ -1,6 +1,8 @@
 """
 Main training program
 """
+from os import path, makedirs
```
nashirj, line 4: sorry to pep8 u again but can you just `import os` and then do `os.blah` everywhere you use it?

chess-AI/train.py
nashirj: I guess it could all just be in a `load_dataset` function that does this

chess-AI/train.py
fshcat: did this one

nashirj (approved):

Issue #90, opened 2022-01-14
Write moveDirect and moveAlongEdges functions in movementFunction.ino


LSaber36:

LSaber36:
This requires issues #91 and #185 to be addressed first

LSaber36:
This is how the primitive functions should be used with the handler function:
```
// statusCodeResult should be declared in the scope of the calling function 
uint8_t statusCodeResult;

.
.
.


// As an example, I'll use moveStraight here
// The same thing needs to be done with moveDiagonal

statusCodeResult = moveStraight(motor, endRow, endCol)
// Checks status; if not SUCCESS, call statusCodeHandler
// Returns false if statusCodeHandler returns false, otherwise, continues
if (statusCodeResult != SUCCESS && !statusCodeHandler(statusCodeResult))
{
  return false;
} 

// All other function calls here
...

// If we reach end of function, we don't need Arduino to request a retransmission
return true;
```

PR #188, opened 2022-03-07
Implement error handling for movement functions

adenm-10:


firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -172,6 +172,9 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
   int *currentMotorPos;
   int startCol, startRow;
 
+  // Assigned at the end of each movement function to be checked for error handling
```
LSaber36, line 175: A better way to phrase this would be to say that you're assigning the value using the return from each movement primitive call. Remember, the call executes the function and the result is put in place of the function wherever it was called.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -245,6 +248,14 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
   *currentMotorPos += (eighthStepsPerPulse * numSteps);
 
   return SUCCESS;
+
+  // Error code handling
```
LSaber36, line 252: Though this is the correct way we want to do this, we're not going to do that in this PR. This PR mainly serves to give us the handler function. We're going to actually write the `moveDirect` and `moveAlongEdges` functions in a PR for issue #90. In other words, please remove the assignment on line 254 and the if statement starting on line 255.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -344,4 +356,12 @@ uint8_t moveDiagonal(int endCol, int endRow)
   currPositionY += (eighthStepsPerPulseY * numStepsX);
 
   return SUCCESS;
+
+  // Error code handling
```
LSaber36, line 360: Same as above.

firmware/arduino/chessboard/motorDriver.ino
LSaber36: I just noticed this as well, but to clarify, these types of calls will be done in `moveDirect` and `moveAlongEdges`, not in either `moveStraight` or `moveDiagonal`

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -259,6 +270,7 @@ uint8_t moveDiagonal(int endCol, int endRow)
   int eighthStepsPerPulseX, eighthStepsPerPulseY;
   int i;
   int startRow, startCol;
+  uint8_t statusCodeResult;
```
LSaber36, line 273: Same as above

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -172,6 +172,9 @@ uint8_t moveStraight(uint8_t motor[], int endCol, int endRow)
   int *currentMotorPos;
   int startCol, startRow;
 
+  // Assigned at the end of each movement function to be checked for error handling
+  uint8_t statusCodeResult;
```
LSaber36, line 176: The comment above and this declaration need to be done in `moveDirect` and `moveAlongEdges` in `movementFunctions.ino`

LSaber36: This is based on where we're calling the primitives, which I explain in the next comment.

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -37,3 +37,37 @@ void centerPiece()
     // Turn electromagnet off
     digitalWrite(ELECTROMAGNET, LOW);
 }
+
+// The handler function that acts on different error codes output by movement primitives
```
LSaber36, line 41: A better way to phrase comments before a function definition would be to start with a verb, and explain what it does, also with any constraints it has. That might look like this: 
```
// Handles behavior based on different error codes passed in
// Must take in an error code to work properly, returns true by default
```

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -37,3 +37,37 @@ void centerPiece()
...
+    else if (status == INVALID_ALIGNMENT)
+    {
+        home();
+        return true;
+    }
+    else if (status == INVALID_ARGS)
+    {
+        return false;
+    }
+}
```
LSaber36, line 73: The reason you're getting that error and the slight optimization I have fortunately have the same fix. Notice that if you pass in a `status` that isn't checked for, the function can't actually return a bool, which is what it's expecting, that's where your error is coming from. The slight optimization is this. See how it's mostly returning true except for one case? You can make the function return true by default at the bottom, and if you ever meet the conditions to return false, you do that before you arrive at the return true statement.

LSaber36: This is the structure I'm going for:
```
bool foo ()
{
  if (condition1)
  {
    action1
  }
  else if (condition2)
  {
    action1
  }
  else if (condition3)
  {
    action1
    return false;
  }

  return true;
}
```

LSaber36: This way, you always return something and avoid the repeated `return true;` statements :)

PR #197, opened 2022-03-15
Implement movement functions

LSaber36:


firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+
+  // Make initial move to first position. Move diagonally since it's faster.
+  if (!moveDirect(curCol, curRow, startCol, startRow))
+    return false;
+
+  // Calculate a list of up to 4 new points (including the start point) 
+  // We use previous points to calculate subsequent points, since they're on the same path
+  // Initial position
+  cols[0] = startCol;
+  rows[0] = startRow;
```
nashirj, line 59: Can move this after the
```c
if (absDeltaX <= 2  &&  absDeltaY <= 2)
    return moveDirect(startCol, startRow, endCol, endRow);
```

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+  {
+    uint8_t diagDirX2, diagDirY2;
+
+    if (absDeltaX == 0)
+    {
+      // Handle edge case where start position is on the edge of the board, so you must move inward
+      diagDirX = (startCol == (TOTAL_UNITSPACES - 1)) ? -1 : 1;
+
+      subDeltaX = 0;
+      subDeltaY = deltaY - (2 * diagDirY);
```
nashirj, line 77: You already compute `subDeltaY` above, don't need to repeat here.

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+      subDeltaX = 0;
+      subDeltaY = deltaY - (2 * diagDirY);  
+      diagDirX2 = -diagDirX;
+    }
+    else
+    {
+      // Handle edge case where start position is on the edge of the board, so you must move inward
+      diagDirY = (startRow == (TOTAL_UNITSPACES - 1)) ? -1 : 1;
+
+      subDeltaX = deltaX - (2 * diagDirX);
```
nashirj, line 85: You already compute `subDeltaX` above, don't need to repeat here.

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+
+  // Because of the checks above, we know absDeltaX and absDeltaY can't both be 0 at the same time
+  // Case where we have a strictly vertical or horizontal movement along edges
+  if (absDeltaX == 0  ||  absDeltaY == 0)
+  {
+    uint8_t diagDirX2, diagDirY2;
+
+    if (absDeltaX == 0)
+    {
+      // Handle edge case where start position is on the edge of the board, so you must move inward
```
nashirj, line 73:
```c
// Handle edge case where start position is on the rightmost edge of the board,
// so you must move left for the first diagonal motion. Move right for first
// diagonal in all other cases of strictly vertical motion.
```

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+      diagDirX = (startCol == (TOTAL_UNITSPACES - 1)) ? -1 : 1;
+
+      subDeltaX = 0;
+      subDeltaY = deltaY - (2 * diagDirY);  
+      diagDirX2 = -diagDirX;
+    }
+    else
+    {
+      // Handle edge case where start position is on the edge of the board, so you must move inward
```
nashirj, line 82:
```c
// Handle edge case where start position is on the topmost edge of the board,
// so you must move down for the first diagonal motion. Move up for first
// diagonal in all other cases of strictly horizontal motion.
```

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+      subDeltaX = deltaX - (2 * diagDirX);
+      subDeltaY = 0; 
+      diagDirY2 = -diagDirY;
+    }
+
+    // Add diagonal movement
+    cols[1] = cols[0] + diagDirX;
+    rows[1] = rows[0] + diagDirY;
+
+    // Add straight movement
```
nashirj, line 94: Add: `Note, at least one of subDeltaX or subDeltaY is equal to 0`

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+
+    // Add diagonal movement, with flipped direction
+    cols[3] = cols[2] + diagDirX2;
+    rows[3] = rows[2] + diagDirY2;
+
+    // 3 added points, plus the initial point
+    numPoints = 4;
+  }
+  // Case where we're moving a cached piece to the graveyard from a capture
+  else if (startCol % 2 == 1  ||  startRow % 2 == 1)
```
nashirj, line 106: Comparing to 1 is redundant here, can just do `(startCol % 2 || startRow % 2)`, since any nonzero number evaluates to true.

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+  {
+    subDeltaX = deltaX - diagDirX; 
+    subDeltaY = deltaY - diagDirY;
```
nashirj, line 109: These are only used once, so can just do them in line if you prefer:
```c
cols[1] = cols[0] + (deltaX - diagDirX);
```
and similar for `rows[2]`

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+
+    // Add diagonal movement
+    cols[3] = cols[2] + diagDirX;
+    rows[3] = rows[2] + diagDirY;
+
+    // 3 added points, plus the initial point
+    numPoints = 4;
+  }
+  // Case where we have a knight or graveyard movement
+  else if ((absDeltaX == 4  &&  absDeltaY == 2)  ||  (absDeltaX == 2  &&  absDeltaY == 4))
```
nashirj, line 127: I gave you a bad conditional to check here. We want to check for cases where there is any amount of horizontal motion and `absDeltaY == 2` or any amount of vertical motion and `absDeltaX == 2`. So slightly more general than what is currently here. To that end, this should instead be
```c
if (absDeltaX == 2  ||  absDeltaY == 2)
```

If either is equal to 2, the other must be greater than equal to 2, because otherwise we would have hit one of the two cases we checked above
```c
if (absDeltaX <= 2  &&  absDeltaY <= 2)
if (absDeltaX == 0  ||  absDeltaY == 0)
```

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+  {
+    if (absDeltaX == 2)
+      subDeltaX = 0;
+    else
+      subDeltaY = 0;
+
+    // Add diagonal movement
+    cols[1] = cols[0] + subDeltaX;
+    rows[1] = rows[0] + subDeltaY;
```
nashirj, line 136: should be `diagDir*` not `subDelta*`

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+    rows[2] = rows[1] + subDeltaY;
+
+    // Add diagonal movement
+    cols[3] = cols[2] + diagDirX;
+    rows[3] = rows[2] + diagDirY;
+
+    // 3 added points, plus the initial point
+    numPoints = 4;
+  }
+  // Default case, where all other movements should
```
nashirj, line 149: This comment is a little unclear

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,259 @@
...
+        return false;
+    }
+    else if (cols[i] == cols[i-1])
+      statusCodeResult = moveStraight(xMotor, cols[i], rows[i]);
+      if (statusCodeResult != SUCCESS && !statusCodeHandler(statusCodeResult))
+        return false;
+    else
+      statusCodeResult = moveDiagonal(cols[i], rows[i]);
+      if (statusCodeResult != SUCCESS && !statusCodeHandler(statusCodeResult))
+        return false;
```
nashirj, line 192: You need brackets for these lines

firmware/arduino/chessboard/movementFunctions.ino
LSaber36: I know this, I wanted it to be explicit but yeah we don't really need it.

firmware/arduino/chessboard/movementFunctions.ino
LSaber36: We could do this, but I would like to keep them the way they are since it makes it a bit more readable. (All of the other array assignments are like this)

firmware/arduino/chessboard/movementFunctions.ino
LSaber36: I agree with this, thanks for catching it!

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,255 @@
...
+  // There are 4 points total, but only 3 total for each x/y component
+  uint8_t rows[5], cols[5];
+  int8_t diagDirX, diagDirY;
+  int8_t deltaX, deltaY, subDeltaX, subDeltaY, absDeltaX, absDeltaY;
+  uint8_t i, statusCodeResult;
+  uint8_t curCol, curRow;
+  uint8_t numPoints = 0;
+
+  curCol = currPositionX / (stepsPerUnitSpace * 8);
+  curRow = currPositionY / (stepsPerUnitSpace * 8);
```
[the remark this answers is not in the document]
LSaber36: It can't hurt, I'll add it.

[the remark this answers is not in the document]
LSaber36: I'm actually not sure about this. Is it really necessary to check twice? Might it be better to check in moveAlongEdges and not in the primitives? To my knowledge, nothing should be able to make it unaligned, especially not between moveAlongEdges and a primitive call.

adenm-10 (approved):

LSaber36:
I just realized this, but when a movement function leads to returning false, the electromagnet doesn't turn off.

firmware/arduino/chessboard/movementFunctions.ino
```diff
@@ -1,72 +1,255 @@
...
+
+  // If target point is equal to the start point
+  if (deltaX == 0  &&  deltaY == 0)
     return true;
+  
+  // Make initial move to first position. Move diagonally since it's faster.
+  if (!moveDirect(curCol, curRow, startCol, startRow))
+    return false;
+
+  // Calculate a list of up to 4 new points (including the start point)
```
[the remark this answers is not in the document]
nashirj: This should either say up to 4 new points (excluding the start point) or up to 5 new points.

firmware/arduino/chessboard/movementFunctions.ino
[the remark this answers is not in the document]
nashirj: On second thought, since this is checked in the primitives, it doesn't necessarily need to be done here, but it would save some computation (building the list, etc) in the event that we are misaligned. I think it should still be checked in the primitives since we might get misaligned between primitive function calls but not before ending in moveAlongEdges.

tl;dr: up to you whether you want to add a check to moveAlongEdges.

nashirj (commented):
I can think of two ways to solve the issue with the electromagnet
1) Convert the for loop that iterates over the movements to a while loop. Break from the loop if you can't handle the status (i.e. where we are currently returning false). After the loop, turn off the electromagnet. Return `loop counter variable < numPoints` (if we broke early, it will be less than, so we know that movement failed).
2) If we receive false in the calling function, turn off the electromagnet.

I prefer option 1.

firmware/arduino/chessboard/movementFunctions.ino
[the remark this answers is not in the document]
LSaber36: We should leave it cuz it may get misaligned later on. Though, this'll get removed later whenever we implement the encoder stuff.

Issue #190, opened 2022-03-11
Update AlignAxis info bit handling


nashirj:
Setting the first info bit (e.g. `msg[2]`, where `msg == "~<OpCode><SetZero>000<OpType>"`) to '0' indicates aligning x axis to zero, '1' indicates aligning y axis to zero, indicates aligning x axis to max, indicates aligning y axis to max.

Issue #149, opened 2022-02-10
Add HOME_AXIS and TOGGLE_EM to OpCode enum


nashirj:
We need to add a new HOME_AXIS field to the OpCode enum on Pi and Arduino side.

Arduino will need to add handling of that OpCode in the parser; if the code is received, should call the `homeaxis()` function.
message format will be `"~<HOME_AXIS>0000<move_number>"`

nashirj:
Also want to be able to turn EM on and off.

Message format:
1) Turn on: "~<TOGGLE_EM>1000<move_number>"
1) Turn off: "~<TOGGLE_EM>0000<move_number>"

PR #198, opened 2022-03-15
Update arduino uart scheme

chasesza:
Address #190 and #149

### DIFF
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -52,7 +52,7 @@ enum ErrorCode
   MOVEMENT_ERROR = '4'
 };
 
-enum InstructionTypes
+enum InstructionType
 {
   ALIGN_AXIS = '1',
   SET_ELECTROMAGNET = '2',
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -203,16 +203,16 @@ void loop()
     { 
         // Sends acknowledgement
         sentMessage[0] = currentState;
-        sentMessage[1] = moveCount;
-        sentMessage[2] = errorCode;
+        sentMessage[1] = errorCode;
+        sentMessage[2] = moveCount;
         sendMessageToPi(sentMessage);
 
         makeMove(receivedMessagePtr);
     }
     // Sends move success/error
     sentMessage[0] = currentState;
-    sentMessage[1] = moveCount;
-    sentMessage[2] = errorCode;
+    sentMessage[1] = errorCode;
+    sentMessage[2] = moveCount;
     sendMessageToPi(sentMessage);
   }
 
@@ -220,8 +220,8 @@ void loop()
   if (buttonFlag)
   {
     sentMessage[0] = currentState;
-    sentMessage[1] = moveCount;
-    sentMessage[2] = errorCode;
+    sentMessage[1] = errorCode;
+    sentMessage[2] = moveCount;
     sendMessageToPi(sentMessage);
     buttonFlag = false;
   }
@@ -231,8 +231,8 @@ void loop()
   {
     uartMessageIncompleteFlag = false;
     sentMessage[0] = currentState;
-    sentMessage[1] = moveCount;
-    sentMessage[2] = errorCode;
+    sentMessage[1] = errorCode;
+    sentMessage[2] = moveCount;
     sendMessageToPi(sentMessage);
   }
 }
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -210,6 +210,7 @@ void loop()
         makeMove(receivedMessagePtr);
     }
     // Sends move success/error
+    // These variables can be changed inside the makeMove function
     sentMessage[0] = currentState;
     sentMessage[1] = errorCode;
     sentMessage[2] = moveCount;
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -18,7 +18,7 @@ volatile char messageBuffer2[INCOMING_MESSAGE_LENGTH];
 volatile char * tempCharPtr;
 volatile char * rxBufferPtr = messageBuffer1;
 volatile char * receivedMessagePtr = messageBuffer2;
-volatile char sentMessage[3];
+volatile char sentMessage[3]; // sentMessage[] holds status, errorCode, and moveCount (in that order)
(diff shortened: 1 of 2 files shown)


firmware/arduino/chessboard/uart.ino
```diff
@@ -84,6 +84,15 @@ bool validateMessageFromPi(volatile char * message)
             return false;
         }
     }
+    else if (message[0] == INSTRUCTION)
+    {
+        if (message[2] < ALIGN_AXIS || message[2] > RETRANSMIT)
```
nashirj, line 89: Indexing here is off; for `Instruction` messages, the last char specifies `OpType`. So that would be `message[5]`

firmware/arduino/chessboard/uart.ino
```diff
@@ -131,6 +140,37 @@ bool makeMove(volatile char * message)
             return false;
         }
     }
+    // Move type 3 - special instructions
+    else if (message[0] == INSTRUCTION)
+    {
+        // Align Axis
+        if (message[2] == ALIGN)
```
nashirj, line 147: Same comment about `OpType` indexing

firmware/arduino/chessboard/uart.ino
```diff
@@ -131,6 +140,37 @@ bool makeMove(volatile char * message)
...
+          else if (message[3] == '1')
+            alignAxis(yMotor, ZERO_POSITION);
+          else if (message[3] == '2')
+            alignAxis(xMotor, MAX_POSITION);
+          else if (message[3] == '3')
+            alignAxis(yMotor, MAX_POSITION);
+        }
+
+        // Enable/Disable Electromagnet
+        else if (message[2] == SET_ELECTROMAGNET)
```
nashirj, line 160: Same

firmware/arduino/chessboard/uart.ino
```diff
@@ -131,6 +140,37 @@ bool makeMove(volatile char * message)
...
+        {
+          if (message[3] == '0')
+            digitalWrite(ELECTROMAGNET, LOW);
+          else if (message[3] == '1')
+            ledcWrite(EM_PWM_CHANNEL, PWM_HALF);
+        }
+
+        // Retransmit last message
+        else if (message[2] == RETRANSMIT)
```
nashirj, line 169: Same

firmware/arduino/chessboard/uart.ino
```diff
@@ -131,6 +140,37 @@ bool makeMove(volatile char * message)
...
+        // Align Axis
+        if (message[2] == ALIGN)
+        {
+          if (message[3] == '0')
+            alignAxis(xMotor, ZERO_POSITION);
+          else if (message[3] == '1')
+            alignAxis(yMotor, ZERO_POSITION);
+          else if (message[3] == '2')
+            alignAxis(xMotor, MAX_POSITION);
+          else if (message[3] == '3')
```
nashirj, line 155: `Extra` char is `message[1]`

firmware/arduino/chessboard/uart.ino
```diff
@@ -131,6 +140,37 @@ bool makeMove(volatile char * message)
...
+            alignAxis(yMotor, MAX_POSITION);
+        }
+
+        // Enable/Disable Electromagnet
+        else if (message[2] == SET_ELECTROMAGNET)
+        {
+          if (message[3] == '0')
+            digitalWrite(ELECTROMAGNET, LOW);
+          else if (message[3] == '1')
+            ledcWrite(EM_PWM_CHANNEL, PWM_HALF);
```
nashirj, line 165: Same about `Extra`

firmware/arduino/chessboard/uart.ino
nashirj: I don't know if there is a better way to do this, but if we change the number of `Instruction OpType`s, this line needs to be updated. So currently this is the most efficient way to do it, but it'll break later if we're not careful. Any thoughts on that?

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -193,25 +202,37 @@ void loop()
     if (validateMessageFromPi(receivedMessagePtr))
     { 
         // Sends acknowledgement
-        sendMessageToPi(currentState, moveCount, errorCode);
+        sentMessage[0] = currentState;
+        sentMessage[1] = moveCount;
+        sentMessage[2] = errorCode;
```
nashirj, line 207: See below comment about reordering these fields, and similar below

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -50,6 +52,13 @@ enum ErrorCode
   MOVEMENT_ERROR = '4'
 };
 
+enum InstructionTypes
```
nashirj, line 55: I like this name better than `OpType`, I'll change the docs and the pi code

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -18,6 +18,7 @@ volatile char messageBuffer2[INCOMING_MESSAGE_LENGTH];
 volatile char * tempCharPtr;
 volatile char * rxBufferPtr = messageBuffer1;
 volatile char * receivedMessagePtr = messageBuffer2;
+volatile char sentMessage[3];
```
nashirj, line 21: Please add a comment above explaining what each field in `sentMessage` is. Something liek
```c
// sentMessage[] holds currentState, errorCode, and moveCount (in that order)
```

nashirj (approved):

firmware/arduino/chessboard/chessboard.ino
```diff
@@ -193,25 +202,37 @@ void loop()
...
         // Sends acknowledgement
-        sendMessageToPi(currentState, moveCount, errorCode);
+        sentMessage[0] = currentState;
+        sentMessage[1] = errorCode;
+        sentMessage[2] = moveCount;
+        sendMessageToPi(sentMessage);
 
         makeMove(receivedMessagePtr);
     }
     // Sends move success/error
```
LSaber36, line 212: It would be a good idea to edit the comment to clarify that you're re-setting the variables and sending them out since they have the potential to change from makeMove. Otherwise it just looks like repeated code.

firmware/arduino/chessboard/uart.ino
```diff
@@ -84,6 +84,15 @@ bool validateMessageFromPi(volatile char * message)
             return false;
         }
     }
+    else if (message[0] == INSTRUCTION)
+    {
```
LSaber36, line 88: Please add a comment explaining that you're checking if message[5] is not a valid instruction type

PR #204, opened 2022-03-29
Removed homing and added print statements

LSaber36:
Calling the homing function in `setup()` causes the program to not start if no endstops are hooked up. We may want to test the system without them in debug mode or something, so we'll have the Raspberry Pi send the homing commands over UART.

The print statements are to make it clear that the setup function has executed correctly and is now entering the main loop.

### DIFF
--- a/firmware/arduino/chessboard/chessboard.ino
+++ b/firmware/arduino/chessboard/chessboard.ino
@@ -179,14 +179,14 @@ void setup()
   ledcSetup(EM_PWM_CHANNEL, PWM_FREQUENCY, PWM_RESOLUTION);
   ledcAttachPin(ELECTROMAGNET, EM_PWM_CHANNEL);
 
-  // Homes the EM to the origin and sets currPositionX/Y to 0
-  home();
-
   // Initializes global 2d array `pulsesPerSlope` which is used to define circle paths
   // that are used in the `makeCircle()` function in circleFunction.ino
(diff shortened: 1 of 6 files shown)


nashirj:
Changes needed on Arduino side (@chasesza)

1. need to use extra field in arduino status to send back opcode. need this in addition to move number, since we now have two move number counts, one for moves, and one for instructions
	extra will hold an error code if the state is error, otherwise it'll hold the opcode of the previously sent message
2. need to update INSTRUCTION message format/handling on arduino
	"~{INSTRUCTION}{op_type}{extra}00{move_count}"
3. need to update chars used for instruction types ('A', 'S', 'R')
```py
    # This code indicates Arduino should align an axis
    # Setting first info bit (e.g. msg[2]) to '0' indicates aligning x axis to zero, '1' indicates
    # aligning y axis to zero, '2' indicates aligning x axis to max, '3' indicates aligning y axis
    # to max.
    ALIGN_AXIS = 'A'

    # This code indicates Arduino should set the state of the electromagnet
    # Setting first info bit (e.g. msg[2]) to '0' indicates OFF, '1' indicates ON
    SET_ELECTROMAGNET = 'S'

    # This code indicates Arduino should retransmit last message
    # This code used when a corrupted or misaligned message is received
    RETRANSMIT_LAST_MSG = 'R'
```

nashirj (approved):

PR #208, opened 2022-03-31
Adjust alignAxis for accurate calibration

adenm-10:


firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -147,6 +148,21 @@ void alignAxis(uint8_t motor[], uint8_t alignmentCode)
...
 
+  // Moves EM back off of endstop extreme after a trigger
+  if (motor == xMotor)
+    tempAlignWholeSteps = (alignmentCode == MAX_POSITION) ? MAX_X_ALIGNMENT : MIN_X_ALIGNMENT;
+  else if (motor == yMotor)
+    tempAlignWholeSteps = (alignmentCode == MAX_POSITION) ? MAX_Y_ALIGNMENT : MIN_Y_ALIGNMENT;
+
+  setScale(motor, WHOLE_STEPS);
+
+  for (j = 0; j < tempAlignWholeSteps; j++)
```
LSaber36, line 159: Instead of declaring a new counter, just use `i` again. It's not used after the previous for loop, so it can be re-used here. `i` is also bigger than it needs to be for this particular loop, so tempAlignWholeSteps will fit inside of it and not have an issue.

firmware/arduino/chessboard/motorDriver.ino
```diff
@@ -76,7 +76,8 @@ void alignAxis(uint8_t motor[], uint8_t alignmentCode)
   int eighthStepsPerPulse;
   int i;
   uint8_t endstopPin;
-
+  uint16_t tempAlignWholeSteps, j;
```
nashirj, line 79: You can reuse `i` instead of making a new var, just make sure to reset it to 0 (which you are already doing in the for loop initialization).

nashirj (approved):

PR #211, opened 2022-04-01
Model checkpointing

fshcat:
Saves model during MCTS training according to the value of save_freq. If this is fine save_freq should probably be added to ai_io

### DIFF
--- a/chess-AI/train.py
+++ b/chess-AI/train.py
@@ -1,6 +1,7 @@
 """
 Main training program
 """
+import time
 import chess
 import numpy as np
 import torch
@@ -128,7 +129,7 @@ def create_dataset(games, move_approximator, val_approximator=None, show_dash=Fa
     return dataset
 
 
-def train_on_dataset(dataset, nnet, options, show_dash=False):
+def train_on_dataset(dataset, nnet, options, save=True, show_dash=False):
     """Train with the specified dataset
 
     Attributes:
@@ -214,8 +215,9 @@ def train_on_dataset(dataset, nnet, options, show_dash=False):
         Dashboard.visualize_losses(average_pol_loss, average_val_loss)
 
     # Saves model to specified file, or a new file if not specified.
-    # TODO: Figure frequency of model saving, right now it is after a defined number of epochs.
-    save_model(nnet, options.save_path, options.num_saved_models, options.overwrite)
+    if save:
+        save_model(nnet, options.save_path + "/model_{}".format(int(time.time()*1000)), options.num_saved_models,
+                   options.overwrite)
 
 
 def create_stockfish_dataset(sf_opt, show_dash):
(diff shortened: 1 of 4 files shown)


chess-AI/train.py
```diff
@@ -1,6 +1,7 @@
 """
 Main training program
 """
+import time
```
nashirj, line 4: What is time being used for?

chess-AI/train.py
```diff
@@ -128,7 +129,7 @@ def create_dataset(games, move_approximator, val_approximator=None, show_dash=Fa
     return dataset
 
 
-def train_on_dataset(dataset, nnet, options, show_dash=False):
+def train_on_dataset(dataset, nnet, options, iteration, save=True, show_dash=False):
```
nashirj, line 132: I think `save` should be a field in `options` (and correspondingly the params/ai_io/dashboard). don't worry about updating the dashboard tho

chess-AI/train.py
```diff
@@ -214,8 +215,9 @@ def train_on_dataset(dataset, nnet, options, show_dash=False):
         Dashboard.visualize_losses(average_pol_loss, average_val_loss)
 
     # Saves model to specified file, or a new file if not specified.
-    # TODO: Figure frequency of model saving, right now it is after a defined number of epochs.
-    save_model(nnet, options.save_path, options.num_saved_models, options.overwrite)
+    if save:
+        save_model(nnet, options.save_path + "/model_{}".format(iteration), options.num_saved_models,
```
nashirj, line 219: This might give problems on windows if we use unix style paths. Maybe it would be best to use [pathlib](https://docs.python.org/3/library/pathlib.html) here

chess-AI/train.py
```diff
@@ -235,7 +237,7 @@ def create_stockfish_dataset(sf_opt, show_dash):
     return create_dataset(sf_opt.games, stocktrain_moves, stocktrain_value_approximator, show_dash)
 
 
-def train_on_mcts(nnet, mcts_opt, show_dash=False):
+def train_on_mcts(nnet, mcts_opt, save_freq=10, show_dash=False):
```
nashirj, line 240: Like you said in the PR description, `save_freq` should be part of ai_io (and the params file/dashboard). since we're not working on the dashboard rn don't worry about it adding it there, but pls add it to the params file and update init_params. Can you also add `save_freq` as a field in `mcts_opt`? Then we don't need to update the function signature

chess-AI/train.py
fshcat: the value of save varies between calls, its just internal for making it so MCTS training can control whether the model gets saved or not. save_freq is where the user input happens

chess-AI/train.py
fshcat: hasn't been giving me any problems, torch.save probably uses pathlib internally on string input

PR #217, opened 2022-04-03
Fix serial messaging on pi

nashirj:
This PR implements receiving communications from Arduino. It also guards all comms with the --microcontroller flag. To run code with Arduino interaction, need to run `python cliinterface.py -m`


nashirj:
Got the OK from Lucas over discord, so merging

PR #231, opened 2022-04-09
Firmware testing

adenm-10:


nashirj:
I think I had to do something before you were able to make progress with this but I forget what that was. Do you remember @adenm-10?

PR #239, opened 2022-04-14
Test vs random

JuddBE:


nashirj:
Can I help in any way with this or are you able to close it out Judd?

JuddBE:
> Can I help in any way with this or are you able to close it out Judd?

Thanks for checking but I'm able to work on it. I will get started on it before the meeting on Thursday and finalize it before the next meeting after that. We can discuss how we'd like it to be structured/organized in this upcoming meeting tho if that sounds good, just so we're on the same page with how it's going to look.

chess-AI/model_evaluation.py
```diff
@@ -0,0 +1,185 @@
...
+                    policy,_ = player2.model(get_cnn_input(board))
+                    board.push_uci(policy_converter.find_best_legal_move(policy, board))
+                else:
+                    board.push(player2.select_move(board))
+                if board.is_game_over(claim_draw=True):
+                    if not(board.is_checkmate()):
+                        pointsP1 += 0.5
+                    break        
+        else:
+            while True:
```
ryPattillo, line 120: The following code could probably be extracted out to a method since its mostly duplicated code.

chess-AI/model_evaluation.py
```diff
@@ -0,0 +1,185 @@
...
+                if board.is_game_over(claim_draw=True):
+                    if board.is_checkmate():
+                        pointsP1 += 0.5
+                    pointsP1 += 0.5
+                    break
+    pointsP2 = num_games - pointsP1
+    return pointsP1, pointsP2
+
+def get_elo(points, num_games):
+    expected = 0.5
```
ryPattillo, line 145: These should be constants defined outside of the method.

chess-AI/model_evaluation.py
```diff
@@ -0,0 +1,185 @@
...
+        elos[i] = k_value * (points[i] - expected * num_games)
+        
+    maximum = max(elos)
+    minimum = min(elos)
+    for i in range(length):
+        elos[i] = int(((elos[i] - minimum) / (maximum - minimum)) * (HardStockfishPlayer().elo))
+    return elos
+
+def main():
+    num_games = 30
```
ryPattillo, line 159: Could also be a constant defined outside of the method

chess-AI/.gitignore
```diff
@@ -5,8 +5,13 @@
 # Our first model!
 !model.pt
 
+# BayesianElo program used for computing elo of agents
+BayesianElo/
```
ryPattillo, line 9: Where do we get this program?

chess-AI/model_evaluation.py
```diff
@@ -0,0 +1,185 @@
...
+        # if type(move) is not str:
+        #     move = move['Move']
+        return random.choice(list(board.legal_moves))
+
+    def __str__(self):
+        return "RandomStockfishPlayer"
+        
+
+def evaluate_two_players(player1, player2, num_games=20):
+    white = False
```
ryPattillo, line 92: If white is just what player starts then shouldn't this just be random? `white = randint(0,1)` that way each game it is  just a random chance of either our AI starting or the enemy.

chess-AI/.gitignore
nashirj: https://github.com/ddugovic/BayesianElo

chess-AI/model_evaluation.py
nashirj: I know I said I wouldn't be involved, but I'll chime in here to say I think there should be an equal number of games (can be a parameter that you guys figured out) of black and white games for each opponent. e.g. 5 games as white, 5 games as black vs. each opponent. That way the evaluation is less random since the way we compute the evaluation is the same. What do you guys think?

ryPattillo: That would work also. I think we need to track wins for white vs black. rather than just `wins=[0,0,0]` it should probably be `wins_white=[0,0,0]` and `wins_black=[0,0,0]`

chess-AI/.gitignore
ryPattillo: This should probably be included in a the ReadMe for chess-AI. I am going to work on creating a more detailed readme for chess-AI because we have nothing explaining figshare, or params.json.

ryPattillo (approved):
Looks good

PR #244, opened 2022-04-16
Add files via upload

ColonelToad:
I am having trouble with the NumPy load function and I cannot seem to understand why (https://numpy.org/doc/stable/reference/generated/numpy.load.html), so a code review would be appreciated. Note that since we have to work with a smaller dataset than the original raspberry Turk project we should expect differing accuracy values.


ColonelToad:
Link to collab file; https://colab.research.google.com/drive/1Ou0lmY1C9-ilA05Ma8vLsz9K9FfkfFxw?usp=sharing

nashirj:
Great progress!

nashirj:
Now that we know what the problem is (needing to run the data preprocessing scripts first), I'm going to close this PR until you're ready for it to be reviewed!

Issue #84, opened 2022-01-13
Integrate GUI with game.py


nashirj:
Explore python chess GUI to use in conjunction with game.py

nashirj:
The GUI doesn't need to be able to mouse interaction. Just need it to be able to receive messages and use those to update the current board. For now, just taking user input from command line is sufficient, but we'll need to figure out how to integrate message passing later.

nashirj:
The [Python-Easy-Chess-GUI](https://github.com/fsmosca/Python-Easy-Chess-GUI) could be a good reference that we can modify to suit our purposes

PR #247, opened 2022-04-20
Game loop GUI

RuthShryock:
Might need to reconfigure some of the invalid moves, pawn promotion situation, and ending the game.


RuthShryock:
address #84

nashirj (commented):
I'm leaving you a partial review, I'll continue reviewing gui.py, pieces.py and chessboard.py in the meantime and make a second review for those files.

firmware/raspi/pieces.py
```diff
@@ -0,0 +1,176 @@
+"""
+adapted code from:
+*    Title: chess_tk
+*    Author: Guatam Sharma
+*    Date: May 19, 2016
+*    Availability: https://github.com/saavedra29/chess_tk
+
+"""
```
nashirj, line 8:
```py
"""Module containing helper GUI code for gui.py.

Based on https://github.com/saavedra29/chess_tk"""
```

firmware/raspi/pieces.py
```diff
@@ -0,0 +1,176 @@
...
+        if piece.isupper():
+            color = "white"
+        else:
+            color = "black"
+        piece = SHORT_NAME[piece.upper()]
+    module = sys.modules[__name__]
+    return module.__dict__[piece](color)
+
+class Piece(object):
+    """Chess piece class condainting color and shortname.
```
nashirj, line 37: condainting -> containing

nashirj (commented):
Works really well! One thing I noticed (aside from the comments I left you) is that castling does not work yet for the GUI player. When the king is selected, if castling is available, the square for castling should be highlighted for the king. I attached a screenshot of lichess with a position that has castling legal so you can see what it looks like.

Screen Shot 2022-05-07 at 11 11 14 PM

nashirj:
Let me know if I can help with this PR in any way

firmware/raspi/gui.py
```diff
@@ -0,0 +1,359 @@
...
+    def update_state(self):
+        """Runs simultaneously with the tkinter main loop and checks if a move has been
+        made every 10ms 
+        """
+        made_move = self.game.process(self.players[self.turn])
+        if made_move:
+            self.selected_piece = None
+            self.focused = None
+            self.pieces = {}
+            if (self.turn and self.move_first == "gui") or (not self.turn and self.move_first == "ai"):
```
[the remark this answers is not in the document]
RuthShryock: Yes so this if statement will print the arrows on the turn of the opposing player to the human. If the turn is 1/true and the human moved first, then the opposing player moved and we need to show the arrows. If the turn is 0/false and the AI moved first, then the opposing player moved and we need to show the arrows. Since turn will represent the opposing player when it is true and false (depending on which color the human picked), I wrote this if statement to work with the respective conditions

firmware/raspi/gui.py
```diff
@@ -0,0 +1,353 @@
...
+from util import parse_args
+
+class GUI:
+    """GUI class that creates a visual interface of the board.
+    """
+    pieces = {}
+    selected_piece = None
+    focused = None
+    turn = ""
+    move_first = ""
```
nashirj, line 22: Can you please change this to `human_plays_white_pieces` and make it hold `True` or `False` rather than a string?

firmware/raspi/gui.py
```diff
@@ -0,0 +1,353 @@
...
+                self.show_move(self.game.last_made_move(), self.game.is_white_turn())
+            self.draw_board()
+            self.draw_pieces()
+            self.turn = (self.turn + 1) % 2
+        self.parent.after(10, self.update_state)
+
+    def square_clicked(self, event):
+        """Waits for click on the board and depending on player's turn, executes the move
+        """
+        if not self.turn or (self.turn and self.move_first == "ai"):
```
nashirj, line 108: I think this can be simplified to
```py
(not self.turn) or (not self.human_plays_white_pieces)
```
where `not self.human_plays_white_pieces` corresponds to `self.move_first == "ai"`

firmware/raspi/gui.py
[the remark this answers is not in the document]
nashirj: Sounds good; can you add a brief comment explaining that and update this condition to be:
```py
if (self.turn and self.human_moves_first) or (not self.turn and not self.human_moves_first):
```

firmware/raspi/gui.py
```diff
@@ -0,0 +1,353 @@
...
+ 
+    raise ValueError("Error parsing parameters...")
+
+def player_wants_rematch():
+    """Skeleton method for querying player about rematch.
+    """
+    # TODO: implement
+    return False
+
+def main(): #took out chessboard
```
nashirj, line 326: You can remove this comment

firmware/raspi/gui.py
```diff
@@ -0,0 +1,353 @@
...
+                self.players[self.turn].set_move(self.made_move)
+            self.focus(pos)
+            self.draw_board()
+
+        if self.game.is_game_over():
+            print("Checking if game is over...")
+            # Need to empty move queue to play out final captures, etc (if any) before ending
+            while self.game.board.msg_queue:
+                self.game.process(None)
+            #gui player can no longer make moves
```
nashirj, line 134: Can you please update all your comments to have a space between the # and the beginning of the comment? And please capitalize the first word of the comments.

firmware/raspi/gui.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def setup_game(self):
+        """Sets up the main game variables and then calls the tkinter mainloop
+        which waits for events
+        """
+        params = init_parameters()
+        mode_of_interaction = params["mode_of_interaction"]
+        print(f"\nRUNNING IN {mode_of_interaction.upper()} MODE...\n")
+ 
+        if params["mode_of_interaction"] == "test":
```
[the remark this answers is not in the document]
nashirj: Basically, we don't want to have to worry about cli (or other mode of interaction) stuff in this file, and vice versa in the cli file (don't want to have to worry about gui stuff).

firmware/raspi/pieces.py
```diff
@@ -0,0 +1,228 @@
...
+        return map(board.alpha_notation, allowed_moves)
+
+class King(Piece):
+    """King chess piece.
+    """
+    shortname = "k"
+    canCastle = True
+
+    #def moves_available(self, pos):
+        #return super(King, self).moves_available(pos.upper(), True, True, 1)
```
nashirj, line 91: If not using, can remove.

firmware/raspi/pieces.py
```diff
@@ -0,0 +1,228 @@
...
+                whiteKingHome = False
+            else:
+                blackKingHome = False
+
+        for x, y in possible_moves:
+            newLoc = beginningpos[0] + x, beginningpos[1] + y
+            if board.alpha_notation(newLoc) not in board.occupied(self.color):
+                allowed_moves.append(newLoc)
+
+        # Checks if each rook can castsle
```
nashirj, line 125: Fix spelling please

firmware/raspi/pieces.py
```diff
@@ -0,0 +1,228 @@
...
+                    allowed_moves.append(dest)
+                    collision = True
+        allowed_moves = filter(board.is_on_board, allowed_moves)
+        return map(board.alpha_notation, allowed_moves)
+
+class King(Piece):
+    """King chess piece.
+    """
+    shortname = "k"
+    canCastle = True
```
nashirj, line 88: Please update all camel case variable names to be snake case instead: e.g. `can_castle` instead of `canCastle`.

firmware/raspi/pieces.py
```diff
@@ -0,0 +1,228 @@
...
+        """Returns moves available based on piece
+        """
+        board = self.board
+        _ = self
+        if self.color == "white":
+            startpos, direction, enemy = 1, 1, "black"
+        else:
+            startpos, direction, enemy = 6, -1, "white"
+        allowed_moves = []
+        # Moving
```
nashirj, line 211: The function name is moves_available (which makes it seem like it would return the allowed_moves) but this comment makes it seem like this function is moving the pieces, is that what's happening?

nashirj (commented):
Thanks for working on this, overall looks really good. Left some comments but also have some requests for behavior changes:
- If you click a piece and then click it again without making a move, it should no longer be highlighted.
- The castling behavior doesn't work as expected; the GUI doesn't update the rook movement until after the next player's turn. Can we make it so that if the king makes a castling move, both the king and the rook positions are updated simultaneously?
- About ten moves into the game playing as white (I tried twice), the game stops responding - it's white's turn, and after I click for where I want to go, the game freezes.

firmware/raspi/gui.py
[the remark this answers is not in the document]
RuthShryock: Yes, there is a lot of overlapping code. The assign_human_turn_at_start(), init_parameters(), and main() functions from the cli_interface.py are all implemented in the gui_interface.py, with slight modifications, so we should probably find a way to extract it

firmware/raspi/gui.py
```diff
@@ -0,0 +1,353 @@
+"""Module containing main GUI code to interact with game.py.
+
+Based on https://github.com/saavedra29/chess_tk
+"""
+import random
+import sys
+import tkinter as tk
+from tkinter import simpledialog
```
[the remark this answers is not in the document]
RuthShryock: Simpledialog is in a separate module from tkinter (according to this [SO](https://stackoverflow.com/questions/53057749/importing-stuff-from-tkinter-python)) so it needs to be imported separately I believe

PR #265, opened 2022-09-06
StockFish path is added for WIndows

mplekunov:
Added path to stockfish executable for Windows OS.


nashirj (approved):
Thanks for the change!

